PR #3, opened 2014-05-27
Adds additional QStandardItemModel compatibility.

tonybarbieri:
This pull request will add the appropriate QStandardItemModel calls when rebuilding the model.  beginResetModel and endResetModel will inform any views to wait to update until the model has been rebuilt.  These are also necessary for any proxy models that may be using a ShotgunModel as a source.

I also set to block signals while the rebuild is happening so signals aren't being rapidly sent during a full refresh.

Thanks!

### DIFF
--- a/python/shotgun_model/shotgunmodel.py
+++ b/python/shotgun_model/shotgunmodel.py
@@ -134,7 +134,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
 
         self.__download_thumbs = download_thumbs
 
-        
+
 
     ########################################################################################
     # public methods
@@ -672,7 +672,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
         # is universal across time zones and DST changes.
         #
         # When you are pulling data from the shotgun model and want to convert this unix timestamp
-        # to a *local* timezone object, which is typically what you want when you are 
+        # to a *local* timezone object, which is typically what you want when you are
         # displaying a value on screen, use the following code:
         # >>> local_datetime = datetime.fromtimestamp(unix_time)
         #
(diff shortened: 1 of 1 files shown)


python/shotgun_model/shotgunmodel.py
```diff
@@ -948,6 +958,12 @@ def __rebuild_whole_tree_from_sg_data(self, data):
         root = self.invisibleRootItem()
         self.__populate_complete_tree_r(data, root, self.__hierarchy, {})
 
+        # Turn our signals back on as we have effectively reset the model at
+        # this point and finally call the QStandardItemModel's endResetModel
+        # method.
+        self.blockSignals(False)
```
alandann: Is there a reason this isn't _before_ the re-population of the tree from the data?  It it were then this could all go into `__reset_all_data()` which has now been renamed to the public `clear()` in the latest version of the framework.  The behaviour of rows being added to the model would then be consistent regardless of which code path was taken here: https://github.com/tonybarbieri/tk-framework-shotgunutils/blob/qstandarditemmodel_signals/python/shotgun_model/shotgunmodel.py#L690 - what do you think?

alandann:
Hi Tony,

Thanks for this and sorry it's taken so long to pull it in!  Taking a look at your code change I had one comment/question - basically, should the begin/end be around the whole code block or just the 'reset' part where it clears out the old data?  In more recent versions of the framework this has been made public as the 'clear()' method and I'm thinking it makes more sense to go in there?

What do you think?  Would you still get unwanted signals as the model is populated after it's cleared?

Thanks
Alan

python/shotgun_model/shotgunmodel.py
tonybarbieri: I think that should work!  I don't recall why I had put this here, perhaps because a signal existed to notify when the tree was repopulated?  Honestly I don't recall.

tonybarbieri:
It could be due to the way the model was populated one item at a time?  There was a reason, I just don't recall now exactly what it was.

alandann:
Ok, thanks - I'll leave this open for the moment as we have a separate ticket for this which I've linked you change to.

jfboismenu:
Closing this pull request since this was actually [fixed](https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/23) recently.

PR #4, opened 2014-05-27
Adds entity_ids property and index_from_entity method.

tonybarbieri:
Adds a property "entity_ids" which will return a list containing all of the entity_ids currently in the model.

Adds a method "index_from_entity" which returns a QModelIndex for the QStandardItem found that matches the entity_type and entity_id.  If a QStandardItem is not found, None is returned.

Thanks!

### DIFF
--- a/python/shotgun_model/shotgunmodel.py
+++ b/python/shotgun_model/shotgunmodel.py
@@ -134,10 +134,16 @@ class ShotgunModel(QtGui.QStandardItemModel):
 
         self.__download_thumbs = download_thumbs
 
-        
 
     ########################################################################################
     # public methods
+
+    @property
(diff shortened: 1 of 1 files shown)


alandann:
Hi Tony,

`index_from_entity()` makes sense - I'll pull that one in.  I'm curious what your useage of the `entity_ids` property is though?  I suspect it's perfectly fine but just wanted to check before pulling that in.

Thanks
Alan

tonybarbieri:
Hey Alan,

I used entity_ids to be able to iterate all of the entities contained in the model.  The only data I needed was the entity ids so I created a quick way to get at them.

Best,
Tony

alandann:
Thanks Tony :)

PR #5, opened 2014-05-28
Block signals in ShotgunModel.destroy before resetting data.

tonybarbieri:
Hello!

We need to block signals on the ShotgunModel in the destroy method before we reset the data.  If we don't, downstream proxy models may react to the changing data and we'll hit a race condition which cause some hard crashes.

Thanks!


alandann:
Thanks Tony!

PR #6, opened 2014-05-29
Makes the bundle attribute protected rather than private.

tonybarbieri:
Makes the bundle attribute protected rather than private so it's accessible to subclasses!


alandann:
Thanks for this Tony

PR #7, opened 2014-05-29
Makes generate_display_name method protected rather than private.

tonybarbieri:
Makes generate_display_name protected rather than private so it can be overridden by a subclass!

### DIFF
--- a/python/shotgun_model/shotgunmodel.py
+++ b/python/shotgun_model/shotgunmodel.py
@@ -134,7 +134,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
 
         self.__download_thumbs = download_thumbs
 
-        
+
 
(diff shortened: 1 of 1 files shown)


alandann:
Hi Tony,

I'm curious what you changed in your derived class for this one?  The reason I'm asking is to see if you need to override the entire method or if instead there is a missing method that should be called somewhere in `__generate_display_name()`?

Just thought I'd check before pulling this one in.

Thanks
Alan

tonybarbieri:
Hi Alan,

I believe I wanted to tweak how the display name was displayed and not have the id automatically appended.  I was using the model via a proxy model so duplicates were not necessarily side by side.

Best,
Tony

alandann:
Thanks Tony - merged!

PR #12, opened 2015-03-06
#28557 #28558 Fixes for crashing when browsing loader history quickly. Better structure for thumb storage.

manneohrstrom:
Improved workarounds for a known pyside/qt issue where clearing and resetting of models will result in dangling pointers on the PySide side. This change implements `close()` and `reset()` for the Shotgun Model in the interest of stability - anyone trying to call `reset()` will now get an exception raised instead of a (highly likely) scenario with stability issues or crashes. `close()` is reimplemented in order to avoid the default C++ implementation and instead use the hand crafted recursion based clearing that ensures that all objects both on the C++ and Python side are removed.

Thumbnail urls are restructured to better handle long path name as we were running into issues with `MAX_PATH` on windows. Now S3 urls are more intelligently parsed and distributed in a much more sensible way on the file system.

*Behavioural changes:*
- Thumbnails are now stored in a new structure on disk. (Old thumbnails will remain in the cache.)
- Calling `reset()` on the shotgun model will result in an exception raised.

### DIFF
--- a/python/shotgun_data/sgdata.py
+++ b/python/shotgun_data/sgdata.py
@@ -11,6 +11,7 @@
 import os
 import tank
 import uuid
+import hashlib
 import urlparse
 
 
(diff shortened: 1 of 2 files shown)


python/shotgun_data/sgdata.py
```diff
@@ -272,55 +272,63 @@ def request_thumbnail(self, url, entity_type, entity_id, field):
...
+            # for this url, all we really need is the per-file guid
+            # since this is universally unique
+            s3_guid = path_chunks[1]
+            # Now turn this guid into a tree structure. For a discussion about sensible
+            # sharding methodology, see 
+            # http://stackoverflow.com/questions/13841931/using-guids-as-folder-names-splitting-up
+            #
+            # From the guid, generate paths on the form C1C2/C3C4/full_guid.jpeg
+            # for a million evenly distributed items, this means ~15 items per folder
+            first_folder = str(s3_guid[0:2])
```
jfboismenu-adsk: +1 for sharding :)

manneohrstrom: :smile:

python/shotgun_data/sgdata.py
```diff
@@ -272,53 +273,32 @@ def request_thumbnail(self, url, entity_type, entity_id, field):
...
+
+        # Now turn this hash into a tree structure. For a discussion about sensible
+        # sharding methodology, see 
+        # http://stackoverflow.com/questions/13841931/using-guids-as-folder-names-splitting-up
+        #
+        # From the hash, generate paths on the form C1C2/C3C4/full_hash.jpeg
+        # for a million evenly distributed items, this means ~15 items per folder
+        first_folder = str(hash_str[0:2])
+        second_folder = str(hash_str[2:4])
+        file_name = "%s.jpeg" % hash_str
```
jfboismenu-adsk: If you feel that it's worth making strings smaller at https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/12/files#diff-018892ec714588513fa837ee353c2f40R394, then maybe it would be wise to truncate the first 4 characters of the file name since they are already present in the folder path.

manneohrstrom: Good point as always. I'll tweak it.

PR #13, opened 2015-06-05
#29806 info panel extensions - schema methods and spinner access

manneohrstrom:
Some minor additions to the shotgun utils in order to cater for some of the needs of the info panel app. 
These changes are additive so they are fully backwards compatible.

- added a `get_schema()` method to the async shotgun data retriever. This wraps the `schema_read` api method in the Shotgun API.

- added a `_show_overlay_spinner()` method to the overlay model, allowing deriving classes to show the spinner. Previously, deriving classes could hide the spinner and control the overlay text but now show it. This is useful if you want to derive from the model and do some *additional* computational work and want to pop up the spinner for this extra bit of work.

### DIFF
--- a/python/shotgun_data/sgdata.py
+++ b/python/shotgun_data/sgdata.py
@@ -555,10 +555,7 @@ class ShotgunDataRetriever(QtCore.QThread):
                         finally:
                             os.umask(old_umask)
 
-                        # load the thumbnail into memory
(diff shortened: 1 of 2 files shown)


alandann:
Looks good - I'll need to merge it into my branch but I can do that after it's in master.

PR #14, opened 2015-06-16
#29806 info panel extensions pass 2 - more async data methods. 

manneohrstrom:
- Schema method combines both schema_read and schema_entity_read
- `execute_find()` call in sg data fetcher now supports arbitrary arguments passed to `sg.find()`
- Added delete, update and create methods
- Added method which allows for the execution of an arbitrary method
- Fixed an issue where the wrong request_type was being returned for thumbnails.

### DIFF
--- a/python/shotgun_data/sgdata.py
+++ b/python/shotgun_data/sgdata.py
@@ -54,7 +54,7 @@ class ShotgunDataRetriever(QtCore.QThread):
     #
     # - request_type is a string denoting the type of request
     #   this event is associated with. It can be either "find"
-    #   or "thumbnail"
+    #   "update", "create", "delete" "schema" or "thumbnail"
     #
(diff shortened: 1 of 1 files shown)


manneohrstrom:
@alandann would be awesome if you could cast your eye on these changes.

alandann:
This looks fine - certainly nothing that stands out as being bad.  In my other branch I've refactored things slightly to separate out the worker thread from the execution code (e.g. sg queries, thumbnail download) but I don't think it will be hard to merge this new functionality in...

manneohrstrom:
great! I intentionally tried to stay as close to the existing structure as possible, knowing that you have some changes in the pipe. Hopefully this is just "more of the same" and doesn't introduce anything new that will become an conceptual issue when folding into your branch! Cool, thanks, i'll merge it with master.

PR #15, opened 2015-07-02
#30106 asynchronous loading of thumbnails

manneohrstrom:
Fully backwards compatible. Adds a new `bg_thumbs` flag to the ShotgunModel family;

```
def __init__(self, parent, download_thumbs=True, schema_generation=0, bg_thumbs=False):
         """
         Constructor. This will create a model which can later be used to load
         and manage Shotgun data.
 
         :param parent: Parent object.
         :param download_thumbs: Boolean to indicate if this model should attempt
                                 to download and process thumbnails for the downloaded data.
         :param schema_generation: Schema generation index. If you are changing the format
                                   of the data you are retrieving from Shotgun, and therefore
                                   want to invalidate any cache files that may already exist
                                   in the system, you can increment this integer.
        :param bg_thumbs: If set to True, thumbnails will be loaded in the background.
 
         """
```

Setting the flag to True will request that the sgdata async fetcher not only downloads the thumbnail but also loads it up into a `QImage` object. This means that most I/O will happen in the worker thread rather than in the main UI thread. It seems to lead to much better behaviour, especially when you have lots of images.

When `bg_thumbs` is set to `True`, the method `_populate_thumbnail_image()` is called from the base class instead of the normal `_populate_thumbnail()` method. Code subclassing the Shotgun model needs to implement this `_populate_thumbnail_image()` method if they want to take advantage of the new behaviour.

Here's a comparative test on a slow cache storage - showing the new and the old cache running side by side: https://s3.amazonaws.com/uploads.hipchat.com/21107/193181/u7AisV1yqUxMuxx/ScreenFlow.mov

This also improves some other areas:

- Thumbnails are downloaded faster from S3. Instead of doing a find prior to every S3 download, this is only done on failure.
- Thumbs queued up several times (not an uncommon situation) are only dowloaded once.
- Improved debugging.
- Fixed issue with S3 logic inside attachment entities, causing queries pulling down attachment fields with uploads to always invalidate and refresh
- Fixed an issue with callbacks firing when thumbnails are removed on the server and not returned
- Shotgun connections are now created JIT - for optimal performance of initial thumbnail download tasks

Also please note https://github.com/shotgunsoftware/tk-core/commit/14d6bd5e0ec9d663de7486daf11ea7ee6e14e460 - which memoizes cache location. This was slowing down the caching beause it was being called several times. The default hook calls `self.parent.shotgun.base_url`, causing the sg API to initialize inside each worker thread, causing further slowdowns.

### DIFF
--- a/python/shotgun_data/sgdata.py
+++ b/python/shotgun_data/sgdata.py
@@ -365,7 +365,7 @@ class ShotgunDataRetriever(QtCore.QThread):
 
         return uid
 
-    def request_thumbnail(self, url, entity_type, entity_id, field):
+    def request_thumbnail(self, url, entity_type, entity_id, field, load_image=False):
         """
         Adds a Shotgun thumbnail request to the queue.
 
@@ -378,6 +378,8 @@ class ShotgunDataRetriever(QtCore.QThread):
         :param entity_id: Shotgun entity id with which the thumb is associated.
         :param field: Thumbnail field. Normally 'image' but could also for example
                       be a deep link field such as 'sg_sequence.Sequence.image'
+        :param load_image: If set to True, the return data structure will contain
+                           a QImage object with the image data loaded.
 
         :returns: A unique identifier representing this request. This
                   identifier is also part of the payload sent via the
@@ -390,7 +392,8 @@ class ShotgunDataRetriever(QtCore.QThread):
                 "url": url,
                 "field": field,
                 "entity_type": entity_type,
-                "entity_id": entity_id }
+                "entity_id": entity_id,
+                "load_image": load_image }
         self._queue_mutex.lock()
         try:
             self._thumb_check_queue.append(work)
@@ -553,11 +556,7 @@ class ShotgunDataRetriever(QtCore.QThread):
             self._queue_mutex.lock()
             try:
 
-                if len(self._thumb_check_queue) > 0:
-                    item_to_process = self._thumb_check_queue.pop(0)
-                    item_type = ShotgunDataRetriever._THUMB_CHECK
-
-                elif len(self._sg_requests_queue) > 0:
+                if len(self._sg_requests_queue) > 0:
                     item_to_process = self._sg_requests_queue.pop(0)
                     if item_to_process["action"] == "execute_find":
                         item_type = ShotgunDataRetriever._SG_FIND_QUERY
@@ -576,6 +575,11 @@ class ShotgunDataRetriever(QtCore.QThread):
 
                     elif item_to_process["action"] == "execute_method":
                         item_type = ShotgunDataRetriever._EXECUTE_METHOD
+
+                elif len(self._thumb_check_queue) > 0:
+                    item_to_process = self._thumb_check_queue.pop(0)
+                    item_type = ShotgunDataRetriever._THUMB_CHECK
+
 
                 elif len(self._thumb_download_queue) > 0:
                     item_to_process = self._thumb_download_queue.pop(0)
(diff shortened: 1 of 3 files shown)


python/shotgun_data/sgdata.py
```diff
@@ -690,34 +700,60 @@ def run(self):
...
                 elif item_type == ShotgunDataRetriever._THUMB_DOWNLOAD:
                     # download the actual thumbnail. Because of S3, the url
-                    # has most likely expired, so need to re-fetch it via a sg find
+                    # may have expired - in that case fall back, get a fresh url
+                    # from shotgun and try again
                     entity_id = item_to_process["entity_id"]
                     entity_type = item_to_process["entity_type"]
                     field = item_to_process["field"]
-
-                    sg_data = self.__sg.find_one(entity_type, [["id", "is", entity_id]], [field])
```
manneohrstrom: Avoiding this find() lookup happening for every thumbnail drastically speeds up thumbnail loading! Now instead, the find() lookup only happens when the S3 url doesn't work (likely because of expiry).

manneohrstrom:
@JesseSpielmanFS heads up about this pull request - this resolves a number of the issues we discussed earlier in the week and should hopefully result in better loader performance for you guys.

manneohrstrom:
All right @alandann - I think I have piled enough changes into this PR now. Ready for CR :)

python/shotgun_data/sgdata.py
```diff
@@ -532,6 +535,23 @@ def _get_thumbnail_path(url, bundle):
...
+    def __ensure_sg_connection_initialized(self):
+        """
+        Populates self.__sg unless it has already been set.
+        
+        This allows for a workflow when the sg connection is 
+        created just in time before a sg operation, allowing
+        maximum speed for cache I/O tasks running in the data
+        fetcher (like grabbing cached thumbnails from disk). 
+        """
+        if self.__sg is None:
```
jfboismenu-adsk: One safer way to have this functionality would probably be to have a property _sg that lazily initializes __sg and use self.sg in the code?

python/shotgun_data/sgdata.py
```diff
@@ -594,61 +613,64 @@ def run(self):
...
+                        self._queue_mutex.lock()
+                        try:
+                            self._thumb_download_queue.append(item_to_process)
+                        finally:
+                            self._queue_mutex.unlock()
 
-                if item_type == ShotgunDataRetriever._SG_FIND_QUERY:
+                elif item_type == ShotgunDataRetriever._SG_FIND_QUERY:
                     # get stuff from shotgun
+                    self.__ensure_sg_connection_initialized()
```
jfboismenu-adsk: Definitely have a property that lazily initializes __sg feels the right solution. This is getting copy pasted all over the place and could easily be missed in a branch.

python/shotgun_model/shotgunmodel.py
```diff
@@ -90,7 +90,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
     FILE_VERSION = 21
 
 
-    def __init__(self, parent, download_thumbs=True, schema_generation=0):
+    def __init__(self, parent, download_thumbs=True, schema_generation=0, bg_thumbs=False):
```
jfboismenu-adsk: I feel like the name bg_thumbs is not clear enough. It sounds confusing when considering the the worker thread retrieves thumbnails in the background. Maybe bg_load_thumbs (or anything that implies that it is being not only downloaded but also read from disk by the background thread) would be clearer?

manneohrstrom: good point. I'll go with `bg_load_thumbs`!

manneohrstrom:
Great feedback @jfboismenu-adsk thanks so much. Code updated and merged with master.

PR #17, opened 2015-08-04
For #28106 - Workfiles v2 changes fully merged with master

alandann:
Lots of changes including refactoring/tidying of code and file names and addition of new classes.

- Renamed all .py files to match their contained classes
- Added ShotgunEntityModel from the tk-multi-loader2 app and extended it to provide swatches for Step entities
- Added BackgroundTaskManager - a class that processes a queue of tasks asynchronously using a pool of worker threads.  Tasks are python callable objects (functions, methods, class with __call__ method, etc.) that can be grouped and prioritised and can be dependent on upstream tasks.
- Updated the ShotgunDataRetriever and ShotgunModel (and derived classes) to take an optional BackgroundTaskManager instance in their constructors to allow a single BackgroundTaskManager to be used for all asynchronous work within an app.  If a BackgroundTaskManager is not provided then one is created internally with a maximum thread limit of 1 to mimic the previous behaviour.

Note that the merge was __extremely__ painful so I ended up taking a clean branch from master and manually merging/redoing my changes on top of that.  This should be fully tested though to ensure the new functionality in master still works!


python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        """
+        return self._priority
+
+    def append_upstream_result(self, result):
+        """
+        Append the result from an upstream task to this tasks kwargs.
+
+        :param result:  A dictionary containing the result from an upstream task
+        """
+        if result and isinstance(result, dict):
```
jfboismenu-adsk: Any reason why we do validation yet don't thrown on error? Seems like we should either
- assume the input is valid
or
- validate it and throw on error

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+
+    def run(self):
+        """
+        Perform this task
+
+        :returns:   The result of performing the task
+        """
+        return self._cbl(*self._args, **self._kwargs)
+
+class _WorkerThreadA(QtCore.QThread):
```
jfboismenu-adsk: Is it me or WorkerThreadA and B are both dead code?

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+
+    @property
+    def shotgun_connection(self):
+        """
+        Get a Shotgun connection to use.  Creates a new Shotgun connection if the instance doesn't 
+        already have one.  The connection returned should only be used by the current thread.
+
+        :returns:   The Shotgun connection for this instance in the current thread
+        """
+        if not hasattr(self._threadlocal_storage, "shotgun"):
```
jfboismenu-adsk: Feels like this is a pattern that should be refactored somehow...

jfboismenu-adsk: Also, not thread safe. If a context switch happens after the if, you're going to initialize the threadlocal storage twice.

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        # clear out all storage:
+        self._running_tasks = {}
+        self._available_threads = []
+        self._pending_tasks_by_priority = {}
+        self._tasks_by_id = {}
+        self._group_task_map = {}
+        self._upstream_task_map = {}
+        self._downstream_task_map = {}
+
+        # stop all worker threads and then wait for them to complete:
```
jfboismenu-adsk: Why all the code in comments?

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        self._log("Waiting for %d background threads to stop..." % len(self._all_threads))
+        for thread in self._all_threads:
+            thread.shut_down()
+            #thread.deleteLater()
+            #thread = None
+            #thread.stop(wait_for_completion = True)
+        self._log("Shut down successfully!")
+        self._all_threads = []
+
+    def add_task(self, cbl, priority=None, group=None, upstream_task_ids=None, args=[], **kwargs):
```
jfboismenu-adsk: The signature of this method can hide subtle bugs. Invoking the method like this `add_task(cb, upstream_taks_ids=[1])` will silently fail if the cb itself takes a kwargs as well or will fail later and not at the point of invocation of the method.

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+            self._downstream_task_map.setdefault(us_task_id, set()).add(new_task.uid)
+
+        self._log("Added Task %s to the queue" % new_task)
+
+        # and start the next task:
+        self._start_tasks()
+
+        return new_task.uid
+
+    def add_pass_through_task(self, priority=None, group=None, upstream_task_ids=None, **kwargs):
```
jfboismenu-adsk: Same argument applies here as in add_task method. Can hide subtle bugs.

jfboismenu-adsk: If this kind of task is meant for synchronization, maybe add_synchronization_task would be a better name. Are there any other uses for this task type?

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        if task is None:
+            return
+
+        self._log("Stopping Task %s..." % task)
+        self._stop_tasks([task], stop_upstream, stop_downstream)
+        self._log(" > Task %s stopped!" % task)
+
+    def stop_task_group(self, group, stop_upstream=True, stop_downstream=True):
+        """
+        Stop all tasks in the specified group from running.  If any tasks are already running then the will
```
jfboismenu-adsk: then they will complete

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+    def shut_down(self):
+        """
+        Shut down the task manager.  This clears the task queue and gracefully stops all running
+        threads.  Completion/failure of any currently running tasks will be ignored.
+        """
+        self._log("Shutting down...")
+        self._can_process_tasks = False
+
+        # clear out all storage:
+        self._running_tasks = {}
```
jfboismenu-adsk: Looks like you should call stop_tasks here instead.

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+            if task:
+                tasks_to_stop.append(task)
+        del self._group_task_map[group]
+        self._stop_tasks(tasks_to_stop, stop_upstream, stop_downstream)
+
+        self._log(" > Task group %s stopped!" % group)
+
+    def stop_all_tasks(self):
+        """
+        Stop all currently queued or running tasks.  If any tasks are already running then the will
```
jfboismenu-adsk: then they will

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        # copy the task list as we'll be modifying it:
+        tasks_to_stop = list(tasks_to_stop)
+        # and make sure we only stop each task once!!
+        stopped_task_ids = set([task.uid for task in tasks_to_stop])
+
+        while tasks_to_stop:
+            task_to_stop = tasks_to_stop.pop(0)
+
+            # get the up & downstream tasks to also stop depending on the flags:
+            if stop_upstream and task_to_stop.uid in self._upstream_task_map:
```
jfboismenu-adsk: These two ifs feel like they should be refactored.

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+            self._remove_task(task_to_stop)
+
+    def _get_worker_thread(self):
+        """
+        Get a worker thread to use.  
+
+        :returns:   An available worker thread if there is one, a new thread if needed or None if the thread
+                    limit has been reached.
+        """
+        if self._available_threads:
```
jfboismenu-adsk: Looking at the code, it seems this sometimes will be called by a thread that adds a task (typically the main thread I presume) and sometimes it's going to be a background thread when it is done with it's current task. Unless I'm mistaken, this looks like it's not thread-safe at all.

jfboismenu-adsk: With that being said, if I'm wrong and any interaction is done from the main thread, it might be ok for this class to not be thread safe, but it should definitely be documented as such.

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        """
+        # start tasks until we fail to start one for whatever reason:
+        started = True
+        while started:
+            started = self._start_next_task()
+
+    def _start_next_task(self):
+        """
+        Start the next task in the queue if there is a task that is startable and there is an 
+        available threads to run it.
```
jfboismenu-adsk: available thread

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        # we need a thread to do the work with:
+        thread = self._get_worker_thread()
+        if not thread:
+            # looks like we can't do anything!
+            return False
+
+        self._log("Starting task %r" % task_to_process)
+
+        # ok, we have a thread so lets move the task from the priority queue to the running list:
+        self._pending_tasks_by_priority[priority] = (self._pending_tasks_by_priority[priority][:task_index]
```
jfboismenu-adsk: I'm still amazed Python lists don't have any better interface to remove an item at a given position...

jfboismenu-adsk: del self._pending_tasks_by_priority[priority][task_index]?

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        # ok, we have a thread so lets move the task from the priority queue to the running list:
+        self._pending_tasks_by_priority[priority] = (self._pending_tasks_by_priority[priority][:task_index] 
+                                                     + self._pending_tasks_by_priority[priority][task_index+1:])
+        if not self._pending_tasks_by_priority[priority]:
+            # no more tasks with this priority so also clean up the list
+            del self._pending_tasks_by_priority[priority]
+        self._running_tasks[task_to_process.uid] = (task_to_process, thread)
+
+        num_tasks_left = 0
+        for pending_tasks in self._pending_tasks_by_priority.values():
```
jfboismenu-adsk: Speaking from personal preference, but I generally prefer to use itervalues to avoid the creation of a temporary array to iterate over them.

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+        :param task:    The task that completed
+        :param result:  The task result
+        """
+        try:
+            # check that we should process this result:
+            if task.uid in self._running_tasks:
+                self._log("Task %r - completed" % (task))
+
+                # if we have dependent tasks then update them:
+                for ds_task_id in self._downstream_task_map.get(task.uid) or []:
```
jfboismenu-adsk: A quick comment why get could return None would be appreciated. From my understanding, it's because it's possible on of the task that depends on you could have been removed for some reason before and therefore get would return None. Which begs the question, why not .get(task.uid, [])?

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+
+                # emit signal that this task is completed:
+                self.task_completed.emit(task.uid, task.group, result)
+
+                if group_finished:
+                    # also emit signal that the entire group is completed:
+                    self.task_group_finished.emit(task.group)
+        finally:
+            # move this task thread to the available threads list:
+            self._available_threads.append(self.sender())
```
jfboismenu-adsk: Kinda violating encapsulation of the signal's source here. I think it would have been preferable if the QThread was obtained some other way. This will break in the future if for some reason an object that is not the thread invoked the slot.

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+
+                    # emit failed signal for the failed task:
+                    self.task_failed.emit(failed_task.uid, failed_task.group, msg, tb)
+
+                    if group_finished and failed_task.group not in finished_groups:
+                        self.task_group_finished.emit(failed_task.group)
+                        finished_groups.add(failed_task.group)
+        finally:
+            # move this task thread to the available threads list:
+            self._available_threads.append(self.sender())
```
jfboismenu-adsk: Same comment about encapsulation violation.

python/shotgun_data/background_task_manager.py
```diff
@@ -0,0 +1,832 @@
...
+
+        # fist remove from the running tasks - this will stop any signals being handled for this task
+        if task.uid in self._running_tasks:
+            del self._running_tasks[task.uid]
+
+        # find and remove the task from the pending queue:
+        if task.priority in self._pending_tasks_by_priority:
+            for ti, p_task in enumerate(self._pending_tasks_by_priority.get(task.priority, [])):
+                if p_task.uid == task.uid:
+                    self._pending_tasks_by_priority[task.priority] = (self._pending_tasks_by_priority[task.priority][:ti]
```
jfboismenu-adsk: Again, why not a simple del?

python/shotgun_data/background_task_manager.py
alandann: It's intentional to support a couple of different use cases.  The first (simple) case it to just be able to run a task and get the result back - in this case wrapping the result in a dictionary would be overkill and make the calling code less readable, e.g.:

python/shotgun_data/background_task_manager.py
alandann: It's you ;)

https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/17/files#diff-8235e4ee893b7034fccfa2c8c7308402R612

I deliberately left _WorkerThreadB in there though in case someone decides to try that recipe later and just hits the same problems I did!

python/shotgun_data/background_task_manager.py
alandann: Do you mean centralized?

python/shotgun_data/background_task_manager.py
alandann: Ah, that should be cleaned up - thanks for finding that

python/shotgun_data/background_task_manager.py
alandann: This was actually a deliberate choice to make the interface as simple to use as possible.  I think your example would actually work - the kwargs passed to the task function would just be empty as the upstream_task_ids named argument would be absorbed by the add_task method itself.

The scenario where it would fail is if your task function took the same named parameter as add_task (e.g. priority or group) but this would either fail on invocation of add_task (as there would be multiple named parameters with the same name) or when the task is run (as the named parameter wouldn't be passed through).

Having said all that, it does look a little asymmetric now that args is passed in as a list (the function didn't originally provide support for this) so changing kwargs to a dictionary would probably be fine - I think as long as the task function itself is 'normal' python (e.g. task_func(*args, **kwargs)) then it should be fine.

I'll try making this change and see what it looks like

python/shotgun_data/background_task_manager.py
alandann: Yeah, a use case I have is to add a task that basically acts as a defer mechanism - there are probably other uses as well.  I found myself adding empty tasks in a couple of places for different reasons which led to adding this in the task manager instead.

python/shotgun_data/background_task_manager.py
alandann: fixed

python/shotgun_data/background_task_manager.py
alandann: good spot - changed

python/shotgun_data/background_task_manager.py
alandann: fixed

python/shotgun_data/background_task_manager.py
alandann: not sure what you mean?

python/shotgun_data/background_task_manager.py
alandann: All task processing is handled on the main thread as all task completion/failure signals are processed there but I will update the comments to be clear that the task manager itself is not thread safe.

python/shotgun_data/background_task_manager.py
alandann: fixed

python/shotgun_data/background_task_manager.py
alandann: hmm, not sure why I wrote that the way I did..!  Not only is it horrid, it's also completely unnecessary as I could just do:

```python
self._pending_tasks_by_priority[priority].remove(task_to_process)
```

python/shotgun_data/background_task_manager.py
alandann: do you know what, I didn't even know there was an itervalues() - thanks!

python/shotgun_data/background_task_manager.py
alandann: I think this may have been left over from previous code - changed

python/shotgun_data/background_task_manager.py
alandann: I'm not sure this is necessarily bad..  I could have the signal pass the thread as an argument but that feels like it would be unnecessary duplication as there is already a prescribed mechanism for retrieving the sender and the slot is _only_ intended to be triggered from a thread (it wouldn't make sense otherwise).  I could also have a lookup from the task to the thread but that would also break if the overarching logic changed.  What I will do is rename the slot and update the doc string so it's clear that it is connected to a signal coming from a thread.

python/shotgun_data/background_task_manager.py
alandann: Same answer ;)

python/shotgun_data/background_task_manager.py
alandann: fixed

python/shotgun_data/background_task_manager.py
alandann: hmm, actually changing this makes the interface to add tasks a little clunky.  

For example, this:

```python
self._task_manager.add_task(task_do_something,
        upstream_task_ids = [task_check_status],
        priority = 10,
        group = 123,
        url = url,
        entity_type = entity_type, 
        entity_id = entity_id,
        field = field,
        load_image = load_image)
```

Turns into this:

```python
self._task_manager.add_task(task_do_something,
        upstream_task_ids = [task_check_status],
        priority = 10,
        group = 123,
        task_kwargs = {"url":url,
                "entity_type":entity_type, 
                "entity_id":entity_id,
                "field":field,
                "load_image":load_image})
```
All those extra quotes are pretty nasty!  Although the second version is definitely safer and clearer which parameters are going to the task function and which aren't...  What do you think?

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: Maybe it's just because I just woke up 30 minutes ago, but I still don't see why we silently reject results that are not dictionnaries. It seems like that would be a user error when append_upstream_result and should be flagged appropriately by the system.

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: Yup. We're already doing it in Tank, bundles and now this. It's not a lot of code, mind you, but it we ever find an issue with it at some point (behaving badly with QtThreads at some point wouldn't surprise me for example) we're going to have to fix it in a lot of place.

That's definitely not a high prioriry and is a fairly simple pattern to search for if the issue occurs and fix. Also, the fact that each object has it's own cache seems wasteful, since it means we can end up with multiple instances of the shotgun handle in each thread. It's not like there's a lot of data either attached to that instance however.

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: Sorry, I should have been clearer, there was a voluntary typo in that example : `upstream_taks_ids`, missing an `s`. Because the caller made a mistake in his parameter name, it means that `add_task` didn't get the actual task ids and the caller will not get the expected behaviour because your kwargs swallowed his bad parameter and forwarded them to the task callable.

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: I meant it as "8 lines of copy-pasted code with only variable name changes should be refactored and turned into a method so that maintaining them would be easier"

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: Also, but if you already know the position, del is faster.

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: iteritems and iterkeys as well :)

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: Great idea.

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: Same good idea.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,723 @@
...
+        Convenience and compatibility method for quick and easy synchrnous thumbnail download.
+        This will retrieve a shotgun thumbnail given a url - if it already exists in the cache,
+        a path to it will be returned instantly. If not, it will be downloaded from Shotgun,
+        placed in the standard cache location on disk and its path will be returned.
+
+        This is a helper method meant to make it easy to port over synchronous legacy
+        code - for a better solution, we recommend using the thumbnail retrieval
+        that runs in a background thread.
+
+        Bcause Shotgun thumbnail urls have an expiry time, make sure to only
```
jfboismenu-adsk: Because

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,723 @@
...
+        if not os.path.exists(path_to_cached_thumb):
+
+            # create folders on disk
+            bundle.ensure_folder_exists(os.path.dirname(path_to_cached_thumb))
+
+            # download using standard core method. This will ensure that
+            # proxy and connection settings as set in the SG API are used
+            sgtk.util.download_url(bundle.shotgun, url, path_to_cached_thumb)
+
+            # modify the permissions of the file so it's writeable by others
```
jfboismenu-adsk: I'm wondering if we should have a discussion in the engineering meeting around permissions for stuff in the cache, due to client having NDAs.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,723 @@
...
+        Stop the retriever thread
+        """
+        if not self._task_manager:
+            return
+
+        if self._owns_task_manager:
+            # we own the task manager so we'll need to completely shut it down before
+            # returning
+            self._task_manager.shut_down()
+            self._task_manager = None
```
jfboismenu-adsk: A comment in the method (or maybe if the class docstring?) mentioning the data retriever can't be restarted would be nice, since it's behaviour is contrary to the BgTaskManager.

python/shotgun_data/background_task_manager.py
alandann: I'm not actually a fan of having this separate instance here and would much rather it just use the thread-specific instance from the core api...  

That said, this is here to match the pattern from the previous implementation of the data retriever that had its own shotgun instance.  The reason it moved to here is specifically to avoid each data retriever having a shotgun instance per thread (so num shotgun connections would have equaled num data retrievers x num threads) - however, I think it probably makes more sense to drop this completely and just use the shotgun instance from core.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,723 @@
...
+        """
+        Executes a Shotgun find query asyncronously.
+
+        This method takes the same parameters as the Shotgun find() call.
+
+        :param *args:       args to be passed to the Shotgun find() call
+        :param **kwargs:    Named parameters to be passed to the Shotgun find() call
+        :returns:           A unique task id representing this request. 
+        """
+        if not self._task_manager:
```
jfboismenu-adsk: This pattern of
if not task_manager -> return
add_task
return str(task_id)
is repeated 7 times. I sure hope there's no missing piece of logic in there or I'd have to copy paste the solution 7 times. ;)
self._add_task?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,723 @@
...
+        # read in details about all fields
+        sg_field_schema = self._shotgun_connection.schema_read(project)
+
+        # and read in details about all entity types
+        sg_type_schema = self._shotgun_connection.schema_entity_read(project)
+
+        # need to wrap it in a dict not to confuse pyqt's signals and type system
+        return {"action":"schema", "fields":sg_field_schema, "types":sg_type_schema}
+
+    def _task_execute_find(self, *args, **kwargs):
```
jfboismenu-adsk: Having _task_execute_crud that takes in the crud method as te first param would cut down on code and documentation required.

python/shotgun_data/background_task_manager.py
alandann: I think this is one of those cases where splitting the common code out into a separate method would actually make the code less readable and you end up with more boilerplate around the call to the method to process the result that it ends up being a similar number of lines that get duplicated!

python/shotgun_data/background_task_manager.py
alandann: I did consider that but the code is cleaner with remove (not having to keep track of the position) and the performance would be extremely small

python/shotgun_data/background_task_manager.py
alandann: Yeah, I tend to use iteritems a lot but I'd never come across the other two before - learn something new every day :)

python/shotgun_data/shotgun_data_retriever.py
alandann: Fixed

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,723 @@
...
+
+                else:
+                    # download from sg
+                    url = sg_data[field]
+                    sgtk.util.download_url(self._shotgun_connection, url, thumb_path)
+
+            if thumb_path:
+                # now we have a thumbnail on disk, either via the direct
+                # download, or via the url-fresh-then-download approach
+                # the file is downloaded with user-only permissions
```
jfboismenu-adsk: Would read better if it said something like "The file is download with user-only permissions, so modify the permissions of the file so it's writable by others."

python/shotgun_data/shotgun_data_retriever.py
alandann: yeah but it would be a crappy api ;)

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,723 @@
...
+
+    def _on_task_failed(self, task_id, group, msg, tb):
+        """
+        Slot triggered when a task fails for some reason
+
+        :param task_id: The id of the task that failed
+        :param msg:     The error/exception message for the failed task
+        :param tb:      The stack trace of the exception raised by the failed task
+        """
+        if group != self._bg_tasks_group:
```
jfboismenu-adsk: When can that happen?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,723 @@
...
+            # ignore - it isn't our task!
+            return
+
+        print "TASK %s FAILED:" % task_id
```
jfboismenu-adsk: Would it make sense to print this using bg_manager.log?

python/shotgun_data/shotgun_data_retriever.py
alandann: heh - yeah and there are bugs in some of those as well :(
let me tidy this up...

python/shotgun_data/shotgun_data_retriever.py
jfboismenu-adsk: it's an internal api tough. ;)

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,258 @@
...
+from sgtk.platform.qt import QtGui, QtCore
+
+from .shotgun_model import ShotgunModel
+from .util import get_sg_data, get_sanitized_data
+from .ui import resources_rc
+
+class ShotgunEntityModel(ShotgunModel):
+    """
+    This model represents the data which is displayed inside one of the treeview tabs
+    on the left hand side.
```
jfboismenu-adsk: That sounds awefully specific ;)

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,258 @@
...
+        :param entity_type:         The type of the entities that should be loaded into this model.
+        :param filters:             A list of filters to be applied to entities in the model - these 
+                                    will be passed to the Shotgun API find() call when populating the 
+                                    model
+        :param hierarchy:           List of Shotgun fields that will be used to define the structure 
+                                    of the items in the model.
+        :param fields:              List of Shotgun fields to populate the items in the model with.
+                                    These will be passed to the Shotgun API find() call when populating
+                                    the model.
+        :param parent:              Parent object.
```
jfboismenu-adsk: That's the Qt parent object, right?

python/shotgun_data/shotgun_data_retriever.py
alandann: I think clarity wins here - having separate methods for each shotgun call makes the code easier to read and debug, especially since the callstack you'd get from one of these methods is going to be minimal (as it's in a thread) - having an exception from a _task_execute_crud is going to be a lot harder to figure out the cause than if it says _task_execute_update for example.

python/shotgun_data/shotgun_data_retriever.py
alandann: this was from the original code (quite a bit of the data retriever functionality is) but I'll tidy it up anyway.

python/shotgun_data/shotgun_data_retriever.py
alandann: I'll add a comment ;)

alandann: actually, there is already a comment - the task doesn't belong to the task group (the task manager could be processing tasks from many different sources).

python/shotgun_data/shotgun_data_retriever.py
alandann: that shouldn't be in there - oops!

python/shotgun_model/shotgun_entity_model.py
alandann: yes and fixed

python/shotgun_model/shotgun_entity_model.py
alandann: heh - copy paste - fixed

python/shotgun_data/shotgun_data_retriever.py
alandann: Good point - this follows the pattern of the original code and would have been better if there was a stop() that went with start() and a separate shut_down() to do clean-up.  Unfortunately changing it now could lead to existing code failing unexpectedly (not shutting down threads is very bad in Qt) so I'll add comments and raise an exception in start() if there is no BackgroundTaskManager (because it was previously stopped).

python/shotgun_data/shotgun_data_retriever.py
alandann: Typically the cache is either in the users home directory (so user permissions will already apply) or in a shared network location where the client can set the permissions on the containing directory if needed.  We've had this conversation a few times in the past and agreed that as permissions can be set up for the cache directories as needed (via a core hook) that this isn't a major issue...

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # now walk up the tree and look for an entity in the fields of the 
+        # parent items:
+        while current_item:
+            field_data = get_sanitized_data(current_item, self.SG_ASSOCIATED_FIELD_ROLE)
+            field_value = field_data.get("value")
+            if (field_value 
+                and isinstance(field_value, dict) 
+                and "id" in field_value 
+                and "type" in field_value):
```
jfboismenu-adsk: Just throwing that out there, I'm curious to know what you think: on multi line ifs I prefer putting the closing ): on the next line aligned with the if, this way it's more obvious where the conditional ends and where the indented block starts. That being said, don't bother going through the code to fix this. :)

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,258 @@
...
+        # first, if this is a leaf item then it will represent an entity:
+        sg_data = current_item.get_sg_data()
+        if sg_data:
+            entities.append(sg_data)
+            current_item = current_item.parent()
+
+        # now walk up the tree and look for an entity in the fields of the 
+        # parent items:
+        while current_item:
+            field_data = get_sanitized_data(current_item, self.SG_ASSOCIATED_FIELD_ROLE)
```
jfboismenu-adsk: The innards of the while loop contains non trivial code that is also copy pasted in get_entity. Would be nice to centralize the functionality.

jfboismenu-adsk: I'm also curious why you don't look at item.get_sg_data first.

jfboismenu-adsk: Ah, just read the doc for get_sg_data, seems like it never returns data for intermediary nodes. Is that why?
Still, I feel like this implementation would feel cleaner and more future proof. Am I missing something?
```
entities = []
current_item = item

while current_item:
   sg_data = self.get_entity(current_item)
   if sg_data:
      entities.append(sg_data)
    current_time = current_item.parent
```

python/shotgun_model/shotgun_entity_model.py
alandann: I don't personally don't like parenthesis sitting on a line on their own - it looks a bit sad...
  ):

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,258 @@
...
+
+        # get the associated field data with this node
+        field_data = get_sanitized_data(item, self.SG_ASSOCIATED_FIELD_ROLE)
+        # get the full sg data for this node (leafs only)
+        sg_data = get_sg_data(item)
+
+        # {'name': 'sg_sequence', 'value': {'type': 'Sequence', 'id': 11, 'name': 'bunny_080'}}
+        field_value = field_data["value"]
+
+        if isinstance(field_value, dict) and "name" in field_value and "type" in field_value:
```
jfboismenu-adsk: I wish we had simple predicates we could invoke to test these kind of things.

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+        Get the default icon for the specified entity.
+
+        :param sg_entity:   A Shotgun entity dictionary for the entity to get the
+                            icon for.
+        :returns:           A QIcon for the entity if available.  For Step entities, a
+                            swatch representing the step colour is returned.  If no
+                            icon is available for the entity type then None is returned
+        """
+        if sg_entity.get("type") == "Step":
```
jfboismenu-adsk: This feels specific to the Task entity. Would a ShotgunTaskModel make sense?

python/shotgun_model/shotgun_entity_model.py
alandann: I think this code evolved over time but you are right, get_entities should just call get_entity - I'll change it.  The overhead is pretty low having that extra get_sg_data call for each level of the hierarchy and the code is more consistent.

python/shotgun_model/shotgun_entity_model.py
alandann: the code needed isn't that bad and tbh there are so many variations that you'd end up having to read up on the predicate every time you came across it to remind yourself what it checks...

python/shotgun_model/shotgun_entity_model.py
alandann: nah, I think that would be over specialisation - the model is supposed to work for any entity type but it just happens that the icon for Step is dynamic (a swatch).  You may also use this model for another entity type but include Step somewhere in the hierarchy so the icon would be needed there anyway.

python/shotgun_model/shotgun_model.py
```diff
@@ -90,37 +89,29 @@ class ShotgunModel(QtGui.QStandardItemModel):
...
 
-        :param parent: Parent object.
-        :param download_thumbs: Boolean to indicate if this model should attempt
-                                to download and process thumbnails for the downloaded data.
-        :param schema_generation: Schema generation index. If you are changing the format
-                                  of the data you are retrieving from Shotgun, and therefore
-                                  want to invalidate any cache files that may already exist
-                                  in the system, you can increment this integer.
-        :param bg_load_thumbs: If set to True, thumbnails will be loaded in the background.
+        :param parent:              Parent object.
```
jfboismenu-adsk: Same deal here. Parent model?

python/shotgun_data/shotgun_data_retriever.py
jfboismenu-adsk: Fair point. I hadn't considered the tracebacks.

python/shotgun_data/background_task_manager.py
jfboismenu-adsk: Yeah, I just wrote what I had in mind and it turns out that the boiler plate code needed to get the parameters in and out is not worth it. Too bad.

python/shotgun_data/shotgun_data_retriever.py
jfboismenu-adsk: I should have been clearer: how can the retriever receive a message from another group?

python/shotgun_data/shotgun_data_retriever.py
jfboismenu-adsk: I think it's great actually, just maybe add it as logging, so we can pick it up in the future when we do proper logging?

jfboismenu-adsk:
And I'm done! Over to you.

alandann:
ok, I think everything is addressed...  I'm going to make one last change so that the task manager and data retriever use the Shotgun connections from core before merging...

alandann:
Any last comments or am I ok to merge..?

manneohrstrom:
I'm taking a quick look now!

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,716 @@
...
+        # and append the file name
+        # all sg thumbs are jpegs so append extension too - some url forms don't have this.
+        cache_path_items.append("%s.jpeg" % path_chunks[-1])
+
+        # join up the path
+        path_to_cached_thumb = os.path.join(*cache_path_items)
+
+        return path_to_cached_thumb
+
+    # default individual task priorities
```
manneohrstrom: What's the priority order? I am guessing higher means "more important"?

To be consistent with the current implementation (and I think this behaviour also makes sense), I think priorities should  tweaked to this - does that make sense?

```
# default individual task priorities
# (higher value means more important)

# thumbnail checks are local disk checks and very 
# fast and are always carried out before any shotgun calls
_CHECK_THUMB_PRIORITY           = 50

# the sg schema is often useful to have as early on as possible
# sometimes other shotgun operations also need the shotgun schema
# (and it's typically also cached) so this call has a higher prio than the
# rest of the sg calls
_SG_DOWNLOAD_SCHEMA_PRIORITY    = 40

# next in priority order are all our standard shotgun calls
_EXECUTE_METHOD_PRIORITY        = 30
_SG_DELETE_PRIORITY             = 30
_SG_CREATE_PRIORITY             = 30
_SG_UPDATE_PRIORITY             = 30
_SG_FIND_PRIORITY               = 30

# thumbnails are downloaded last
_DOWNLOAD_THUMB_PRIORITY        = 20

```

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,716 @@
...
+
+    def _task_download_thumbnail(self, thumb_path, url, entity_type, entity_id, field, load_image, **kwargs):
+        """
+        Download the thumbnail for the specified entity type, id and field.  This downloads the
+        thumbnail into the thumbnail cache directory and returns the cached path.
+
+        If thumb_path already contains a path then this method does nothing and just returns the path 
+        without further checking/work.
+
+        :param thumb_path:  Path to an existing thumbnail or None.
```
manneohrstrom: sorry this is most likely just my brain being very jetlagged, but where is `thumb_path` coming from? it's not contained in the dictionary passed to `add_task`

manneohrstrom: is it daisy-chained from the output of _task_check_thumbnail task? How does that work?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -0,0 +1,716 @@
...
+                                                                "entity_type":entity_type, 
+                                                                "entity_id":entity_id,
+                                                                "field":field,
+                                                                "load_image":load_image})
+
+        # all results for requesting a thumbnail should be returned with the same id so use
+        # a mapping to track the 'primary' task id:
+        self._thumb_task_id_map[dl_task_id] = check_task_id
+
+        return str(check_task_id)
```
manneohrstrom: are we casting the task ids to strings to be more backwards compatible? (old version of api returned guid strings) I am worried that it might be easy to miss doing this cast - wondering if we should just return the ints without casting? It's a major version bump so it's ok with things breaking, so maybe better to be consistent between the return values here and the behaviour of the new task manager API - so that they work the same way?

Also, quick question - previously the typical pattern was to stash these task guids as keys in a dict to then process things later - will everything be guaranteed to be unique with the new setup as well - even with ints? Could there be a case where you somehow could end up with two task ids being the same somehow? I am guessing no but thought it's worth just asking!

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,249 @@
...
+import sgtk
+from sgtk.platform.qt import QtGui, QtCore
+
+from .shotgun_model import ShotgunModel
+from .util import get_sg_data, get_sanitized_data
+from .ui import resources_rc
+
+class ShotgunEntityModel(ShotgunModel):
+    """
+    A model that contains a hierarchy of Shotgun entity data and sets the icon for each item
```
manneohrstrom: very useful!

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,249 @@
...
+                                    These will be passed to the Shotgun API find() call when populating
+                                    the model.
+        :param parent:              Parent QObject.
+        :param download_thumbs:     Boolean to indicate if this model should attempt
+                                    to download and process thumbnails for the downloaded data.
+        :param schema_generation:   Schema generation index. If you are changing the format
+                                    of the data you are retrieving from Shotgun, and therefore
+                                    want to invalidate any cache files that may already exist
+                                    in the system, you can increment this integer.
+        :param bg_load_thumbs:      If set to True, thumbnails will be loaded in the background.
```
manneohrstrom: again, with the v3 bump, i think this should be default to True in v3 - it's weird that it's false, right? I just had to make it false to be backwards compatible when the change was added in v2, but I think you typically want this to be happening  - it's a good default behaviour and a big performance boost.

So i think for v3 of the API, `bg_load_thumbs` should default to True everywhere.

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,249 @@
...
+        :param bg_task_manager:     Background task manager to use for any asynchronous work.  If
+                                    this is None then a task manager will be created as needed.
+        """
+        self._entity_icons = {}
+        self._step_swatch_icons = {}
+
+        # make sure fields is valid:
+        fields = fields or []
+        # for backwards compatibility, make sure certain fields are added:
+        fields = list(set(fields + ["image", "sg_status_list", "description"]))
```
manneohrstrom: again, do we need backwards compatibility? I partially see the bump to v3 as a chance to clean some parts of sg utils up a bit, so maybe this sort of thing could be skipped?

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,249 @@
...
+                # create the new icon from this resource:
+                icon = QtGui.QIcon(QtGui.QPixmap(icon_path))
+            self._entity_icons[entity_type] = icon
+
+        return QtGui.QIcon(icon) if icon else None
+
+    def get_entities(self, item):
+        """
+        Get entities for the current item by traversing up the tree and pulling 
+        entity information from each item if possible
```
manneohrstrom: For the parent levels, what fields are returned? is it name/type/id? I think an example here would be useful, for example for a configuration with `Asset` and hierarchy `[project, asset_type, code]`

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,249 @@
...
+            item_entity = self.get_entity(current_item)
+            if item_entity:
+                entities.append(item_entity)
+            current_item = current_item.parent()
+
+        return entities
+
+    def get_entity(self, item):
+        """
+        Get the Shotgun entity details for the specified model item.
```
manneohrstrom: Again, I think an example here would be helpful to illustrate what this method does that `item.get_sg_data()` doesn't!

Also, maybe it's weird that it returns different things. This is something I sometimes tend to avoid in my code.

So if you call this on a intermediate node for a model for `Shot`s with hierarchy `[project, sg_sequence, code]` you would get back `{type: Sequence, id: 1234, name: foo}` but if you call it for a model for `Sequence`s with hierarchy `[project, code]` (e..g sequence is a leaf node) you would potentially get more fields - and perhaps more importantly `code` rather than `name`. I wonder if this should always return name, type and id for all nodes regardless if they are leaves or not. Would that make it easier to use in the calling code? or maybe even just type and id? (but then again it's often useful to have a name)

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,249 @@
...
+        field_value = field_data.get("value")
+        if (field_value 
+            and isinstance(field_value, dict) 
+            and "id" in field_value 
+            and "type" in field_value):
+            return field_value
+
+        return None
+
+    def async_refresh(self):
```
manneohrstrom: should this just be called `refresh()`? Or do we really want to emphasize that it's async? Feels like in the context of the sg utils fw, the refresh is pretty much always async, so maybe ok to just call it refresh?

python/shotgun_model/shotgun_entity_model.py
manneohrstrom: +1 for what alan is saying - the code is already simple - adding a bunch of "helper" functions may just complicate things.

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # {'name': 'sg_sequence', 'value': {'type': 'Sequence', 'id': 11, 'name': 'bunny_080'}}
+        field_value = field_data["value"]
+
+        if isinstance(field_value, dict) and "name" in field_value and "type" in field_value:
+            # this is an intermediate node which is an entity type link
+            entity_icon = self._get_default_thumbnail(field_value)
+            if entity_icon:
+                # use sg icon!
+                item.setIcon(QtGui.QIcon(entity_icon))
```
manneohrstrom: _get_default_thumbnail already returns a QIcon. Do we need to run it through the copy constructor? Is it a GC thing? Maybe this is being overly picky (sorry) but a very small comment could help for future maintainers.

python/shotgun_model/shotgun_entity_model.py
```diff
@@ -0,0 +1,249 @@
...
+            # this is a leaf node!
+            entity_icon = self._get_default_thumbnail(sg_data)
+            if entity_icon:
+                # use sg icon!
+                item.setIcon(QtGui.QIcon(entity_icon))
+                found_icon = True
+
+        # for all items where we didn't find the icon, fall back onto the default
+        if not found_icon:
+            item.setIcon(QtGui.QIcon(self._default_icon))
```
manneohrstrom: Would this code be simpler if _get_default_thumbnail just always returned an icon, never `None` and just returned the default icon in the cases where it didn't find another icon?

manneohrstrom:
Ok I added some comments - take it with a pinch of salt, I am quite jetlagged, brain on three quarter speed at best :) Hopefully there is something useful in there. Overall I think it looks great!

python/shotgun_data/shotgun_data_retriever.py
alandann: It just subscribes to signals emitted by the task manager which can be shared by other objects - I'll update the comment

python/shotgun_data/shotgun_data_retriever.py
alandann: Good spot with the SG queries - I think I misread the original code so I've replaced all the _SG_FIND/CREATE/... priorities with a single _SG_CALL_PRIORITY and updated the comments as you suggested - makes it much clearer.  Thanks

python/shotgun_data/shotgun_data_retriever.py
alandann: it is - I've updated the comments where the tasks are created

python/shotgun_data/shotgun_data_retriever.py
alandann: For this merge I'm going to leave the casting from int to string and back but I agree that it would be better if this wasn't happening - it is purely for backwards compatibility where user code that uses the data retriever may break if an int is returned (I had this happening somewhere at some point I think).

The task ids are guaranteed to be unique within the scope of a single BackgroundTaskManager so the only scenario it would break would be if an instance subscribed to two different BackgroundTaskManager instances with the same slot and didn't validate the sender - not a  pattern that makes much sense.  Using GUIDS just seemed a little overkill and (imo) harder to read/scan output when debugging

python/shotgun_model/shotgun_entity_model.py
alandann: straight from the loader ;)

python/shotgun_model/shotgun_entity_model.py
alandann: This one I completely agree with - I'll change it for the merge

python/shotgun_model/shotgun_entity_model.py
alandann: not sure the comment or the code makes sense - I'll remove it and see what breaks!

python/shotgun_model/shotgun_entity_model.py
alandann: I've updated the comment to be more descriptive - as you point out the information returned will vary depending on the structure of the model.  Limiting it to just type, id and name would be too restrictive I think as it's important to return as much information as possible - typically you want to access the fields you set up in the constructor.

python/shotgun_model/shotgun_entity_model.py
alandann: It certainly would - that code is unnecessarily convoluted!  I think those icon copies were when I was debugging the crash and probably a red herring - I'll remove them and clean up the code

python/shotgun_model/shotgun_entity_model.py
alandann: Probably not needed - see comment below!

python/shotgun_model/shotgun_entity_model.py
alandann: I'm going to leave it as is for the merge as it's consistent with everywhere else but I agree - refresh() would work fine

alandann:
Thanks Manne, I'm going to get this merged into master with the changes that address your comments so at least then others can start using it/working on it..!

manneohrstrom:
This was pushed into https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/21 which was then subsequently merged with master and then updated with documentation. There is therefore nothing here to merge.

PR #20, opened 2015-10-20
#30748 Improved cache logic in shotgun model

manneohrstrom:
This tweaks the messaging and logic around how caches are structured and handled. This change changes the logic for how caches are laid out on disk and old cache files will no longer be reachable. The change addresses several advanced use cases:

- The thumbnail flag is now part of the cache hashing, meaning that shotgun models with and without thumbnails will cache into different locations on disk. This addresses a known issue where the sg model is run in a thread. previously, a `QPixmap: It is not safe to use pixmaps outside the GUI thread` would be reported from the thread in the case the implementation hit a cache which had thumbnails stored. By separating the cache based on the thumb flag, caches are guaranteed to be 'clean' from thumb data whenever the no thumbs flag is used.

-  The `schema_generation` parameter is now cached with the path rather than previously with the file content. This fixes a subtle inconvenience where previously, if you had two versions of an app running, these could "compete" for a cache file, one writing it out only for it to be overwritten by another. The situation would also arise if you happened to have two different apps calling the same query with same params, but having two schema generation markers. Both these scenarios are plausible. 

- The seed parameter is a runtime, data centric parameter and has been moved down into the "runtime" portion of the hash, resulting in cleaner hash file layout when using seed based hashing.

- Improved doc strings.

- Improved exceptions.

### DIFF
--- a/python/shotgun_model/shotgunmodel.py
+++ b/python/shotgun_model/shotgunmodel.py
@@ -22,7 +22,14 @@ from .util import get_sanitized_data, get_sg_data, sanitize_qt, sanitize_for_qt_
 
 from tank.platform.qt import QtCore, QtGui
 
+class ShotgunModelError(tank.TankError):
+    """Base class for all shotgun model exceptions"""
+    pass
 
+class CacheReadVersionMismatch(ShotgunModelError):
+    """Indicates that a cache file is incompatible with this code"""
+    pass
+    
 
 class ShotgunModel(QtGui.QStandardItemModel):
     """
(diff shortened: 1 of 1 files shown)


python/shotgun_model/shotgunmodel.py
```diff
@@ -422,17 +433,42 @@ def _load_data(self, entity_type, filters, hierarchy, fields, order=None, seed=N
...
+        # the schema generation is used for advanced implementations 
+        # See constructor docstring for details.
+        # 
+        # bg_load_thumbs is hashed so that the system can cache 
+        # thumb and non-thumb caches independently. This is because
+        # as soon as you start caching thumbnails, qpixmap will be used
+        # internally by the serialization and this means that you get 
+        # warnings if you try to use those caches in threads. By keeping
+        # caches separate, there is no risk that a thumb cache 'pollutes'
+        # a non-thumb cache.
```
jfboismenu-adsk: Does this mean that in an app toggling thumbs off would force a reload of the content of the view?

manneohrstrom: no no - in that case, you would run the model with thumbnails enabled (set in the constructor of the model), then the view would decide whether to show thumbs or not. the model would keep the thumbs around all the time.

PR #23, opened 2015-12-08
For #32981, rebuild slowness in workfiles

jfboismenu:
- The model now advertises that it is going to be reset, so any listeners can ignore batch updates or empty their caches. This speeds up tremendously how the tree can be cleared when observers are listening to the model since they can temporarily disconnect from the model to avoid a notifications storm or clear caches of QPersistentModelIndex like the Hierarchical. This goes hand in and with this other [pull request](https://github.com/shotgunsoftware/tk-framework-qtwidgets/pull/8).

- Rewrote how we build the hierarchy from a breadth-first to a depth-first approach.

Here are the performance gain we've had
5000 tasks
8.59 vs 3.53 => 2.43x speed gain
10000 tasks
16.84 vs 6.91 => 2.43x speed gain
15000 tasks
60.71 vs 9.88 => 6.14x speed gain
20000 tasks
90.22 vs 13.18 => 6.84x speed gain
25000 tasks
164.13 vs 20.07 => 8.18x speed gain

What is also great to see is that the time it takes to process more nodes is pretty much linear compared to the old approach which didn't scale nicely.

### DIFF
--- a/python/shotgun_model/shotgun_model.py
+++ b/python/shotgun_model/shotgun_model.py
@@ -922,6 +922,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
             # check if anything has been deleted or added
             ids_from_shotgun = set([ d.get("id") for d in sg_data ])
             ids_in_tree = set(self.__entity_tree_data.keys())
+
             removed_ids = ids_in_tree.difference(ids_from_shotgun)
 
             if len(removed_ids) > 0:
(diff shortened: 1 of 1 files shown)


python/shotgun_model/shotgun_model.py
```diff
@@ -1164,141 +1133,135 @@ def __process_thumbnail_for_item(self, item):
...
+        children = node["children"]
+        for child_key in sorted(children.keys(), cmp=lambda x,y: cmp(x.lower(), y.lower())):
+            child = children[child_key]
+            node["item"].appendRow(child["item"])
+            self.__realize_parent_child_hierarchy_r(child)
 
-            # now we got the object set up. Now start calling custom methods:
+    def __populate_complete_tree(self, sg_data):
+        """
+        Generate tree model data structure based on Shotgun data.
```
manneohrstrom: I think it would be cool to add a few sentences here to go through the general approach - explain why you refactored it into a two pass approach like this (for performance). Just some notes to help future readers of the code understand why this particular approach was used!

python/shotgun_model/shotgun_model.py
```diff
@@ -1164,141 +1133,135 @@ def __process_thumbnail_for_item(self, item):
...
+            of ShotgunModelItem to be appended as rows of 'item'.
+        """
+        # If this is a leaf, there are no children.
+        if "children" not in node:
+            return
 
-                # and also populate the id association in our lookup dict
-                self.__entity_tree_data[ sg_item.get("id") ] = item
+        children = node["children"]
+        for child_key in sorted(children.keys(), cmp=lambda x,y: cmp(x.lower(), y.lower())):
```
manneohrstrom: I found this comment helpful in the prev code.
```
# process values in alphabetical order by name, case insensitive
```

manneohrstrom:
cr looks good as far as i can tell! Very minor comments.

PR #24, opened 2015-12-09
Fixed sample code for the documentation on the data retriever

benoit-leveau:
Sample code is wrong.

<pre>
# and start its thread!
self.__sg_data_retriever.start()
</pre>
should be:
<pre>
# and start its thread!
self.__sg_data.start()
</pre>
according to the rest of the code.


jfboismenu:
Thanks Ben!

PR #25, opened 2015-12-09
Added method execute_find_one to the ShotgunDataRetriever class

benoit-leveau:
These requests were available:
* find
* update
* create
* delete
* thumbnail

Adding:
* find_one


manneohrstrom:
Nice one Benoit! Thanks for contributing!

PR #26, opened 2016-01-06
For #34142, no entity names in treeview

jfboismenu:
Do not show the entity type's display name anymore in the treeview. It is now shown via a tooltip for that item. This makes model view less verbose and cleaner to read.
Before:
screen shot 2016-01-06 at 8 39 16 am
After:
screen shot 2016-01-06 at 8 41 39 am

### DIFF
--- a/python/shotgun_model/shotgun_model.py
+++ b/python/shotgun_model/shotgun_model.py
@@ -686,7 +686,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
 
     def _set_tooltip(self, item, sg_item):
         """
-        Called when an item is constructed for the first time.
+        Called when an item is created.
 
         .. note:: You can subclass this if you want to set your own tooltip for the model item. By
             default, the SG_ASSOCIATED_FIELD_ROLE data is retrieved and the field name is used to
@@ -701,12 +701,12 @@ class ShotgunModel(QtGui.QStandardItemModel):
             For example:
             {
                 "type": "Task",
-                "entity": { <-- (1) Tooltip becomes "Asset"
-                    "sg_asset_type: "Character", <-- (2) Tooltip becomes "Asset - Type"
+                "entity": { <-- (1) Tooltip becomes "Asset 'Alice'"
+                    "sg_asset_type: "Character", <-- (2) Tooltip becomes "Link > Type 'Character'"
                     "type": "Asset",
                     "code": "Alice"
                 },
-                "content": "Art" <-- (3) Tooltip becomes "Task"
+                "content": "Art" <-- (3) Tooltip becomes "Task 'Art'"
             }
 
         :param item: Shotgun model item that requires a tooltip.
@@ -718,19 +718,30 @@ class ShotgunModel(QtGui.QStandardItemModel):
 
         if isinstance(sg_item[field], dict) and "type" in sg_item[field]:
             # This is scenario 1 described above.
-            item.setToolTip(get_type_display_name(sg_item[field]["type"]))
+            item.setToolTip(
+                "%s '%s'" % (
+                    get_type_display_name(sg_item[field]["type"]),
+                    item.text()
+                )
+            )
         elif "." in field:
             # This is scenario 2 described above.
             _, sub_entity_type, sub_entity_field_name = field.split(".")
             item.setToolTip(
-                "%s - %s" % (
+                "%s %s '%s'" % (
                     get_type_display_name(sub_entity_type),
-                    get_field_display_name(sub_entity_type, sub_entity_field_name)
+                    get_field_display_name(sub_entity_type, sub_entity_field_name),
+                    item.text()
                 )
             )
         else:
             # This is scenario 3 described above.
-            item.setToolTip(get_type_display_name(sg_item["type"]))
+            item.setToolTip(
+                "%s '%s'" % (
+                    get_type_display_name(sg_item["type"]),
+                    item.text()
+                )
+            )
 
     def _populate_thumbnail(self, item, field, path):
         """
--- a/python/shotgun_model/shotgun_model.py
+++ b/python/shotgun_model/shotgun_model.py
@@ -29,51 +29,51 @@ class ShotgunModelError(tank.TankError):
 class CacheReadVersionMismatch(ShotgunModelError):
     """Indicates that a cache file is incompatible with this code"""
     pass
-    
+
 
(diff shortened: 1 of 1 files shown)


python/shotgun_model/shotgun_model.py
```diff
@@ -669,19 +671,67 @@ def _finalize_item(self, item):
...
 
-        :param item: :class:`~PySide.QtGui.QStandardItem` that is about to be added to the model. 
+        :param item: :class:`~PySide.QtGui.QStandardItem` that is about to be added to the model.
             This has been primed with the standard settings that the ShotgunModel handles.
         """
         # the default implementation does nothing
 
+    def _set_tooltip(self, item, sg_item):
+        """
+        Called when an item is constructed for the first time.
```
manneohrstrom: "For the first time" is confusing - it makes me think that this is only called when the data arrives from shotgun, but it is called every time a new model item is created right? If that's not the case, it should be - imagine a case where 3 different apps share the same cached sg data but want to display it in three different way, therefore needing three different tooltip display implementations!

python/shotgun_model/shotgun_model.py
```diff
@@ -669,19 +671,67 @@ def _finalize_item(self, item):
...
+            default, the SG_ASSOCIATED_FIELD_ROLE data is retrieved and the field name is used to
+            determine which field to pick tooltip information from.
+            - If the field is an entity (e.g. entity), then the display name of that entity's type
+              will be used. (1)
+            - If the field is part of a sub-entity (e.g entity.Asset.sg_asset_type), the display
+              name of the sub-entity's type followed by '-' and the sub-entity's field display name
+              will be used. (2)
+            - If the field is part of an entity and not an entity field(e.g. content), the display
+              name of the entity's type will be used. (3)
+            For example:
```
manneohrstrom: Can you quickly check that this builds correctly with sphinx? I think you need some blank lines in there and `For example::` before the code sample.

python/shotgun_model/shotgun_model.py
```diff
@@ -669,19 +671,67 @@ def _finalize_item(self, item):
...
+            {
+                "type": "Task",
+                "entity": { <-- (1) Tooltip becomes "Asset"
+                    "sg_asset_type: "Character", <-- (2) Tooltip becomes "Asset - Type"
+                    "type": "Asset",
+                    "code": "Alice"
+                },
+                "content": "Art" <-- (3) Tooltip becomes "Task"
```
manneohrstrom: I don't like the UX here. I think if you hover over, it should say `Task Lighting`, not just `Task`. IMO, that looks really confusing! For enum fields I think it should say `Asset, Type Character` with a comma rather than a hyphen.

python/shotgun_model/shotgun_model.py
```diff
@@ -669,19 +671,67 @@ def _finalize_item(self, item):
...
+            }
+
+        :param item: Shotgun model item that requires a tooltip.
+        :param sg_item: Dictionary of the entity associated with the Shotgun model item.
+        """
+
+        data = item.data(self.SG_ASSOCIATED_FIELD_ROLE)
```
manneohrstrom: I think you should do 
```
data = get_sanitized_data(item, self.SG_ASSOCIATED_FIELD_ROLE)
```
Just to be on the safe side.

python/shotgun_model/shotgun_model.py
manneohrstrom: Thinking about this more, I think the case with nested entities should be different:

Say you want to display a list of shots, grouped by a custom field on sequence called `sg_pipeline_name`.

So in the config, it would be `sg_sequence.Sequence.sg_pipeline_name`. 

In Shotgun, the display names are as follows:

- `sg_sequence`: Associated Sequence
- `sg_pipeline_name`: Short Name

I think in this case, the tooltip should say `Associated Sequence > Short Name 'APX'`

This would be consistent with how the table headers for nested fields in the Shotgun web app works.

For deeper (and hence more contrived grouping examples), you could imagine `Sequence > Assignee > Email 'foo@bar.com'` or something. If someone wanted to group the shots by the email address of the person assigned to he sequence :trollface:. I think most importantly, it's good if we can be consistent with shotgun.

manneohrstrom:
CR done! Some suggestions how we can make the tooltips more consistent with the Shotgun web app.

python/shotgun_model/shotgun_model.py
jfboismenu: Isn't [this](https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/master/python/shotgun_model/shotgun_standard_item.py#L87) doing what you are suggesting?

manneohrstrom: ah yes! Sorry! >.<

python/shotgun_model/shotgun_model.py
jfboismenu: I like where you are going with this, but in your example Shotgun would actually show
Associated Sequence > Sequence > Short Name.
So, in my example above, it would become

Asset 'Alicia''
Task 'Art'
Link > Asset > Type 'Character'

python/shotgun_model/shotgun_model.py
```diff
@@ -669,19 +673,87 @@ def _finalize_item(self, item):
...
-            before items appear in the UI. 
+            before items appear in the UI.
 
-        :param item: :class:`~PySide.QtGui.QStandardItem` that is about to be added to the model. 
+        :param item: :class:`~PySide.QtGui.QStandardItem` that is about to be added to the model.
             This has been primed with the standard settings that the ShotgunModel handles.
         """
         # the default implementation does nothing
 
+    def _set_tooltip(self, item, sg_item):
```
jfboismenu: For reference, this is what this block of code gets rendered like: screen shot 2016-01-12 at 11 10 37 am

python/shotgun_model/shotgun_model.py
```diff
@@ -669,19 +673,87 @@ def _finalize_item(self, item):
...
+        """
+        Called when an item is created.
+
+        .. note:: You can subclass this if you want to set your own tooltip for the model item. By
```
jfboismenu: For reference, this is what this block of code gets rendered like: screen shot 2016-01-12 at 11 10 37 am

manneohrstrom: Looks beautiful!

PR #29, opened 2016-02-20
Ticket #34756 - Adds column support to ShotgunModel

robblau:
Extends ShotgunModel and SimpleShotgunModel so that specific fields from the Shotgun query can be used to populate columns in the model.

To use this functionality, simply pass in an array of fields to the _load_data's new optional columns param.

### DIFF
--- a/python/shotgun_model/shotgun_model.py
+++ b/python/shotgun_model/shotgun_model.py
@@ -367,7 +367,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
     ########################################################################################
     # protected methods not meant to be subclassed but meant to be called by subclasses
 
-    def _load_data(self, entity_type, filters, hierarchy, fields, order=None, seed=None, limit=None):
+    def _load_data(self, entity_type, filters, hierarchy, fields, order=None, seed=None, limit=None, columns=None):
         """
         This is the main method to use to configure the model. You basically
         pass a specific find query to the model and it will start tracking
@@ -417,6 +417,10 @@ class ShotgunModel(QtGui.QStandardItemModel):
                           parameter, this can be used to effectively cap the data set that the model
                           is handling, allowing a user to for example show the twenty most recent notes or
                           similar.
+        :param columns:   Normally the model stores the Shotgun data in the SG_DATA_ROLE role.  If columns
+                          is specified, then any leaf row in the model will additionally have columns created
+                          where each item in the row is the value for the corresponding field from columns.
+                          Subclasses can modify this behavior by overriding _get_additional_columns.
 
         :returns:         True if cached data was loaded, False if not.
         """
@@ -432,6 +436,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
         self.__fields = fields
         self.__order = order or []
         self.__hierarchy = hierarchy
+        self.__columns = columns or []
         self.__limit = limit or 0 # 0 means get all matches
 
         # when we cache the data associated with this model, create
@@ -472,6 +477,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
         params_hash.update(str(self.__fields))
         params_hash.update(str(self.__order))
         params_hash.update(str(self.__hierarchy))
+        params_hash.update(str(self.__columns))
 
         # now hash up the filter parameters and the seed - these are dynamic
         # values that tend to change and be data driven, so they are handled
@@ -500,6 +506,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
         self.__log_debug("Hierarchy: %s" % self.__hierarchy)
         self.__log_debug("Fields: %s" % self.__fields)
         self.__log_debug("Order: %s" % self.__order)
+        self.__log_debug("Columns: %s" % self.__columns)
 
         self.__log_debug("First population pass: Calling _load_external_data()")
         self._load_external_data()
@@ -519,6 +526,11 @@ class ShotgunModel(QtGui.QStandardItemModel):
             except Exception, e:
                 self.__log_debug("Couldn't load cache data from disk. Will proceed with "
                                 "full SG load. Error reported: %s" % e)
+
+        # set our headers
+        headers = ["Name"] + self._get_additional_column_headers(self.__entity_type, self.__columns)
+        if headers:
+            self.setHorizontalHeaderLabels(headers)
 
         return loaded_cache_data
 
@@ -567,10 +579,10 @@ class ShotgunModel(QtGui.QStandardItemModel):
             self.__on_sg_data_arrived([])
         else:
             # get data from shotgun - list/set cast to ensure unique fields
+            fields = self.__hierarchy + self.__fields + self.__columns
             if self.__download_thumbs:
-                fields = list(set(self.__hierarchy + self.__fields + ["image"]))
-            else:
-                fields = list(set(self.__hierarchy + self.__fields))
+                fields = fields + ["image"]
+            fields = list(set(fields))
 
             self.__current_work_id = self.__sg_data_retriever.execute_find(self.__entity_type,
                                                                            self.__filters,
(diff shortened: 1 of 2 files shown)


python/shotgun_model/shotgun_model.py
```diff
@@ -841,6 +865,62 @@ def _load_external_data(self):
...
+        will not be called on the return columns.
+
+        This method should return a list of QStandardItems, one for each additional column.
+        The original ShotgunStandardItem is always the first item in each row and should
+        NOT be included in the returned list. Any empty value returned by this method
+        is guaranteed to be treated as an empty list (i.e. you may return None).
+
+        This method is called after _finalize_item.
+
+        :param item: :class:`~PySide.QtGui.QStandardItem` that is about to be added to the model
```
manneohrstrom: Minor: Wondering if the code would be clearer if we indicated that this is the `item` representing the first or the main column of the data we are processing. Maybe call it `primary_item`? (since it carried the payload with all the shotgun data etc?

python/shotgun_model/shotgun_model.py
```diff
@@ -841,6 +865,62 @@ def _load_external_data(self):
...
+        items = []
+
+        if is_leaf and columns:
+            data = get_sg_data(item)
+            for column in columns:
+                # grab the value from the requested field
+                value = data.get(column)
+
+                # set the display role to the string representation of the value
+                column_item = QtGui.QStandardItem(str(value))
```
manneohrstrom: It seems the "primary items" are having their name formatted via the `_generate_display_name()` method. Should this do the same?

python/shotgun_model/shotgun_model.py
```diff
@@ -520,6 +539,11 @@ def _load_data(self, entity_type, filters, hierarchy, fields, order=None, seed=N
                 self.__log_debug("Couldn't load cache data from disk. Will proceed with "
                                 "full SG load. Error reported: %s" % e)
 
+        # set our headers
+        headers = ["Name"] + self._get_additional_column_headers(self.__entity_type, self.__column_fields)
```
manneohrstrom: Nice if we could have "Name" as a constant at the top of the class.

python/shotgun_model/shotgun_model.py
```diff
@@ -841,6 +865,62 @@ def _load_external_data(self):
         """
         pass
 
+    def _get_additional_columns(self, item, is_leaf, columns):
```
manneohrstrom: If I am reading this correctly, we are basically only caching column zero, and this column keeps a full shotgun data chunk representing all the column data. Then at runtime, we "unpack" this by executing `_get_additional_columns()` for each individual row and non-primary column. 

I feel this approach is elegant because it allows for columns to be pretty dynamic in the model, yet at the same time the same cache can be shared and re-used even as columns are treated differently in different cases. Maybe it's premature, but should we be worried about performance? Maybe it's nothing to worry about, but suddenly a majority of the items are created dynamically in python rather than deserialized in directly from disk.

python/shotgun_model/shotgun_model.py
```diff
@@ -520,6 +539,11 @@ def _load_data(self, entity_type, filters, hierarchy, fields, order=None, seed=N
                 self.__log_debug("Couldn't load cache data from disk. Will proceed with "
                                 "full SG load. Error reported: %s" % e)
 
+        # set our headers
+        headers = ["Name"] + self._get_additional_column_headers(self.__entity_type, self.__column_fields)
+        if headers:
```
manneohrstrom: do we need the `if headers` bit? Name is always added.

python/shotgun_model/shotgun_model.py
```diff
@@ -472,6 +489,7 @@ def _load_data(self, entity_type, filters, hierarchy, fields, order=None, seed=N
         params_hash.update(str(self.__fields))
         params_hash.update(str(self.__order))
         params_hash.update(str(self.__hierarchy))
+        params_hash.update(str(self.__column_fields))
```
manneohrstrom: given that none of the column data is cached, is this needed?

python/shotgun_model/shotgun_model.py
```diff
@@ -1291,7 +1374,10 @@ def __realize_parent_child_hierarchy_r(self, node):
         # children are always sorted alphabetically underneath their parent.
         for child_key in sorted(children.keys(), cmp=lambda x,y: cmp(x.lower(), y.lower())):
             child = children[child_key]
-            node["item"].appendRow(child["item"])
+
+            on_leaf_level = not bool(child.get("children"))
```
manneohrstrom:
```python
# child nodes do not have a "children" dictionary key
on_leaf_level = "children" not in child
``` 
would be consistent with code earlier in the method.

manneohrstrom:
This is awesome! Minor CR thoughts and comments. Good to go!! 
Biggest question is whether the on-the-fly creation of columns is a performance hog.

python/shotgun_model/shotgun_model.py
```diff
@@ -267,6 +269,16 @@ def get_entity_type(self):
...
 
+    def get_additional_column_fields(self):
+        """
+        Returns the fields for additional columns and their associated column in the model.
+
+        :returns: A list of (column, field) tuples where column is the column number in the
+            model associated with the additional field.
+        """
+        # column is one greater than the index because of the default initial column
+        return [(i+1, field) for (i, field) in enumerate(self.__column_fields)]
```
manneohrstrom: Sorry one more comment - maybe this should be returned as a dict keyed by index?
So that we could potentially return more data about columns?
I am thinking something along these lines:

```
1: {"field_name": "code", "type": "str", "display_name": "Shot Name"}
2: {"field_name": "entity", "type": "entity_link", "display_name": "Sequence"}
```

Please ignore if you feel I am over thinking things :)

thebeeland: Why not just a list of column-order fields, then?

python/shotgun_model/shotgun_model.py
```diff
@@ -1397,6 +1483,24 @@ def _generate_display_name(self, field, sg_data):
...
+        :param is_leaf: A boolean indicating if the item is a leaf item or not
+
+        :returns: A list of :class:`~PySide.QtGui.QStandardItem`s
+        """
+        # the first item in the row is always the standard shotgun model item,
+        # but subclasses may provide additional columns to be appended.
+        row = [item]
+        additional_columns = self._get_additional_columns(item, is_leaf, self.__column_fields)
+        if additional_columns:
+            row += additional_columns
```
thebeeland: Replace 1499-1501 with:

`row.extend(self._get_additional_columns(item, ...))`

thebeeland: Or even the whole thing with:

`return [item].extend(self._get_additional_columns(item, ...))`

python/shotgun_model/shotgun_model.py
robblau: Switching

python/shotgun_model/shotgun_model.py
robblau: Switching

python/shotgun_model/shotgun_model.py
robblau: Switching

python/shotgun_model/shotgun_model.py
robblau: I ran a test with 10 columns on a model with 100,000 rows and there wasn't any noticeable timing difference between showing the view with or without those columns.  I think we're alright for the kind of calculations we're doing (just pulling values from the SG_DATA dict), although maybe for more complicated calculations this could be an issue.

We talked about it and'll move forward as is for now.

python/shotgun_model/shotgun_model.py
robblau: Switched.

python/shotgun_model/shotgun_model.py
robblau: You're right.  This came from when I was storing more in the cache before the easy path presented itself.  Removed.

python/shotgun_model/shotgun_model.py
robblau: Had that, but ran into a case where children was in child but was an empty dict.  So switched to this.

python/shotgun_model/shotgun_model.py
robblau: A richer interface makes sense.  I didn't add more info to the return value but changing to a dict with keys for column_idx and field.

manneohrstrom: :+1: :cake:

python/shotgun_model/shotgun_model.py
robblau: Did the extend without the return (extend modifies the list but doesn't return it).

PR #33, opened 2016-04-13
For #35801, file save freeze fix

jfboismenu:
Signalling between two different threads in PySide is broken in several versions of PySide. There are very subtle race conditions that arise when there is a lot of signalling  between two threads. Some of these things have been fixed in later versions of PySide, but most hosts integrate PySide 1.2.2 and lower, which are victim of this race condition.

The background task manager is a heavy user of signals between two threads and was sometimes causing deadlocks in many Toolkit applications between PySide's SignalManager and the GIL that would essential freeze the host application.

Therefore, this branch swaps the background task manager's usage of slots and signals for a polling system. The threads inserts results inside a queue and the main thread will flush the queue at regular intervals.

### DIFF
--- a/python/task_manager/background_task_manager.py
+++ b/python/task_manager/background_task_manager.py
@@ -26,9 +26,26 @@ class BackgroundTaskManager(QtCore.QObject):
     Main task manager class. Manages a queue of tasks running them asynchronously through
     a pool of worker threads.
 
-    Note that the BackgroundTaskManager class itself is reentrant but not thread-safe so its methods should only
+    The BackgroundTaskManager class itself is reentrant but not thread-safe so its methods should only
     be called from the thread it is created in. Typically this would be the main thread of the application.
 
+    .. note::
+        Signalling between two different threads in PySide is broken in several versions
+        of PySide. There are very subtle race conditions that arise when there is a lot
+        of signalling between two threads. Some of these things have been fixed in later
+        versions of PySide, but most hosts integrate PySide 1.2.2 and lower, which are
+        victim of this race condition.
+
+        The background task manager does a lot on inter-threads communications and
+        therefore can easily fall pray to these deadlocks that exist within PySide.
+
+        Therefore, the background task manager uses a polling system to process results
+        from the worker threads. The worker threads inserts results inside a queue and
+        the background manager's thread will flush the queue at 100 milliseconds intervals.
+
+        Also note that because there is a 100 milliseconds interval between two polls,
+        it means that the will automatically be a delay between the end of a task
+        and the beginning of another task.
 
     :signal task_completed(uid, group, result): Emitted when a task has been completed.
         The ``uid`` parameter holds the unique id associated with the task,
--- a/python/task_manager/background_task_manager.py
+++ b/python/task_manager/background_task_manager.py
@@ -12,14 +12,13 @@
 Background task manager.
 """
 
+import Queue
 import sgtk
 from sgtk.platform.qt import QtCore
 from sgtk import TankError
 
 from .background_task import BackgroundTask
 from .worker_thread import WorkerThread
-
-import Queue
 
 
 class BackgroundTaskManager(QtCore.QObject):
@@ -46,6 +45,97 @@ class BackgroundTaskManager(QtCore.QObject):
         The ``group`` is the group that has completed.
 
     """
+
+    class _ResultsPoller(QtCore.QObject):
+        """
+        Polls a queue which holds the results posted from the worker threads.
+        """
+
+        # Emitted when a task is completed.
+        task_completed = QtCore.Signal(object, object, object)
+        # Emitted when a task has failed.
+        task_failed = QtCore.Signal(object, object, object, object)
+
+        def __init__(self, parent=None):
+            """
+            Constructor.
+
+            :param parent:  The parent QObject for this thread
+            """
+            QtCore.QObject.__init__(self, parent)
+            # Results queue that will be polled
+            self._results = Queue.Queue()
+            # Create a timer with no interval. This means the timer will be invoked
+            # only when the event queue is empty.
+            self._timer = QtCore.QTimer()
+            self._timer.setInterval(100)
+            # This should be instantiated from the application's thread since we only support
+            # one instance for now.
+            assert(QtCore.QCoreApplication.instance().thread() == self._timer.thread())
+            # Since the event will be invoked as soon as the event queue is empty,
+            # we should process the queue one item at a time to not interfere with the
+            # ui responsivenes
+            self._timer.timeout.connect(self._flush_events)
+            self._bundle = sgtk.platform.current_bundle()
+
+        def start(self):
+            """
+            Start polling for tasks that ended.
+            """
+            self._timer.start()
+
+        def shut_down(self):
+            """
+            Stop polling for tasks that ended.
+            """
+            self._timer.stop()
+
+        def _flush_events(self):
+            """
+            Executes callbacks for each task result.
+            """
+            try:
+                # Get everything until the queue is empty.
+                while True:
+                    cb = self._results.get_nowait()
+                    try:
+                        if len(cb) == 3:
+                            self.task_completed.emit(*cb)
+                        else:
+                            self.task_failed.emit(*cb)
+                    except Exception:
+                        self._bundle.log_exception(
+                            "Exception thrown while reporting completed task."
+                        )
+                        # Do not re-raise, simply process the remaining events.
+            except Queue.Empty:
+                # Queue is empty, nothing to do!
+                pass
+
+        def queue_task_completed(self, worker_thread, task, result):
+            """
+            Called by background threads to notify that a task has completed.
+
+            :param worker_thread: Thread that completed the task.
+            :param task: Task that was completed.
+            :param result: Result produced by the thread.
+            """
+            self._results.put(
+                (worker_thread, task, result)
+            )
+
+        def queue_task_failed(self, worker_thread, task, msg, traceback_):
+            """
+            Called by background threads to notify that a task has failed.
+
+            :param worker_thread: Thread that completed the task.
+            :param task: Task that failed.
+            :param msg: Error message from the task,
+            :param traceback_: Call stack from the task.
+            """
+            self._results.put(
+                (worker_thread, task, msg, traceback_)
+            )
 
     # signal emitted when a task has been completed
     task_completed = QtCore.Signal(int, object, object)# uid, group, result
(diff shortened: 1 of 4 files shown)


python/task_manager/background_task_manager.py
```diff
@@ -1,47 +1,142 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-#
```
jfboismenu: So little needed to be cleaned to be PEP8 compliant that I figure we might as well clean the whole files and be done with it.

python/task_manager/background_task_manager.py
```diff
@@ -1,47 +1,142 @@
...
         the ``message`` is a short error message and the ``traceback_str``
         holds a full traceback.
-        
+
     :signal task_group_finished(group): Emitted when all tasks in a group have finished.
         The ``group`` is the group that has completed.
 
     """
 
+    class _ResultsPoller(QtCore.QObject):
```
jfboismenu: This is the start of the code that was actually changed and not just whitespace changes.

python/task_manager/background_task_manager.py
```diff
@@ -325,15 +429,13 @@ def _get_worker_thread(self):
...
         # in PyQt so WorkerThread is currently preferred - see the notes in the class above for further details
-        thread = WorkerThread(self)
+        thread = WorkerThread(self._results_poller)
         if not isinstance(thread, WorkerThread):
             # for some reason (probably memory corruption somewhere else) I've occasionally seen the above
             # creation of a worker thread return another arbitrary object!  Added this in here so the code
             # will at least continue correctly and not do unexpected things!
             self._bundle.log_error("Failed to create background worker thread for task Manager!")
             return None
-        thread.task_failed.connect(self._on_worker_thread_task_failed)
```
jfboismenu: We no longer connect to the threads, the poller will instead call those callbacks.

python/task_manager/background_task_manager.py
```diff
@@ -417,18 +519,17 @@ def _start_next_task(self):
...
 
         The worker thread instance that completed the task can be retrieved from self.sender()
 
-        :param task:    The task that completed
-        :param result:  The task result
+        :param worker_thread: Thread that completed the task.
+        :param task:          The task that completed
+        :param result:        The task result
         """
-        worker_thread = self.sender()
```
jfboismenu: The sender is now the QTimer, so that's why we're not passing the worker_thread as a parameter.

python/task_manager/worker_thread.py
```diff
@@ -1,36 +1,33 @@
...
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 """
+Worker thread for the background manager.
 """
+
 import traceback
 
-import sgtk
```
jfboismenu: Unused.

python/task_manager/background_task_manager.py
manneohrstrom: nice!

python/task_manager/background_task_manager.py
```diff
@@ -1,47 +1,142 @@
...
-        
+
     :signal task_group_finished(group): Emitted when all tasks in a group have finished.
         The ``group`` is the group that has completed.
 
     """
 
+    class _ResultsPoller(QtCore.QObject):
+        """
+        Polls a queue which holds the results posted from the worker threads.
```
manneohrstrom: Does it make sense to add a blurb here about *why* the heck we are using polling? Ideally with some links to the known pyside fixes that we are working around? I think that could be useful for future reference.

python/task_manager/background_task_manager.py
```diff
@@ -1,47 +1,142 @@
...
+
+            :param parent:  The parent QObject for this thread
+            """
+            QtCore.QObject.__init__(self, parent)
+            # Results queue that will be polled
+            self._results = Queue.Queue()
+            # Create a timer with no interval. This means the timer will be invoked
+            # only when the event queue is empty.
+            self._timer = QtCore.QTimer()
+            self._timer.setInterval(100)
```
manneohrstrom: please make this a constant and put at the top of the file/class.

python/task_manager/background_task_manager.py
```diff
@@ -1,47 +1,142 @@
...
+            # This should be instantiated from the application's thread since we only support
+            # one instance for now.
+            assert(QtCore.QCoreApplication.instance().thread() == self._timer.thread())
```
manneohrstrom: This feels very C++! This would prevent anything from using this outside the main thread right? I don't think that's what we want. The way I see this is that we have a bug fix that helps with stabilty *if* the task mgr is executed from the main thread. Our fix potentially doesn't fix issues if the task manager is run from a worker thread (but may still reduce the frequency of them) but I don't think that means that we should prevent people from using it. Instead, I think we should update the docs with an info/warning box/blurb to highlight the fact that there *may* be issues on some combinations on pyside/python if you run it from a thread.

python/task_manager/background_task_manager.py
```diff
@@ -1,47 +1,142 @@
...
+            self._timer.stop()
+
+        def _flush_events(self):
+            """
+            Executes callbacks for each task result.
+            """
+            try:
+                # Get everything until the queue is empty.
+                while True:
+                    cb = self._results.get_nowait()
```
manneohrstrom: `get_nowait()` is equivalent to just `get()` right? since the queue is of infinite size?

python/task_manager/background_task_manager.py
```diff
@@ -1,47 +1,142 @@
...
+                    try:
+                        if len(cb) == 3:
```
manneohrstrom: this looks odd - is there another way to do it? Or at least add a comment.

python/task_manager/background_task_manager.py
```diff
@@ -325,15 +429,13 @@ def _get_worker_thread(self):
         # create a new worker thread - note, there are two different implementations of the WorkerThread class
         # that use two different recipes.  Although WorkerThreadB is arguably more correct it has some issues
         # in PyQt so WorkerThread is currently preferred - see the notes in the class above for further details
-        thread = WorkerThread(self)
+        thread = WorkerThread(self._results_poller)
```
manneohrstrom: out of curiosity, what's the significance of parenting the thread under the poller rather than under task manager?

python/task_manager/background_task_manager.py
manneohrstrom: Why is this class inside the `BackgroundTaskManager`? I think we should avoid that if we can, even for classes that are only meant to be used in conjunction with a particular other class.

python/task_manager/background_task_manager.py
manneohrstrom: i see, the constructor changed to `def __init__(self, rp, parent=None)` !

I think we should pass the parent too - i suspect there may be issues otherwise:
```
thread = WorkerThread(self._results_poller, parent=self)
```

python/task_manager/worker_thread.py
```diff
@@ -1,36 +1,33 @@
...
-    task_completed = QtCore.Signal(object, object)# task, result
-    # Signal emitted when a task has failed
-    task_failed = QtCore.Signal(object, object, object)# task, message, stacktrace
 
-    def __init__(self, parent=None):
+    def __init__(self, rp, parent=None):
         """
         Construction
 
+        :param rp: Results poller from the background task manager.
```
manneohrstrom: nitpick: clearer if called results_poller rather than rp

python/task_manager/worker_thread.py
```diff
@@ -57,6 +55,7 @@ def shut_down(self):
         """
         Shut down the thread and wait for it to exit before returning
         """
+        self._results_poller = None
```
manneohrstrom: minor: is this a GC thing or is it a way to immediately 'mute' the worker once shut down has been called to ensure no 'straggling tasks' emit signals? May be worth a comment for future readers.

python/task_manager/background_task_manager.py
manneohrstrom: i think we should also add a note that complex chains of linked tasks may suffer from the polling since each task will potentially take 100ms to 'trickle down the chain'.

manneohrstrom:
CR done. I am super excited about this change.  :runner: :watch: :runner: :watch: :runner:

python/task_manager/background_task_manager.py
jfboismenu: Moved out.

python/task_manager/background_task_manager.py
jfboismenu: Actually, this was some leftover from a previous experiment. It isn't necessary anymore because the QTimer is always created with the affinity of the current thread. Therefore, QTimer will always emit timer events in the thread it was created in. This means that creating a background manager in a thread other than the main thread would start sending events to that thread. Obviously, if that thread is not Thread.exec_-ing, then the event loop won't be processed and the results will queue up indefinitely. This was the case with the old system and now this new one as well.

python/task_manager/background_task_manager.py
jfboismenu: get is blocking until something is available. We don't want that, since it would lock the event processing thread.

python/task_manager/background_task_manager.py
jfboismenu: Good catch, thanks!

python/task_manager/worker_thread.py
jfboismenu: I may have inherited Alan's paranoia about cleaning up references to Qt objects. I'll remove it and run my tests to see what happens.

manneohrstrom: paranoia is never a bad thing :) Just curious if it was for a particular reason! :)

python/task_manager/background_task_manager.py
manneohrstrom: ah gotcha! Sorry! #nothingtoseehere...

python/task_manager/background_task_manager.py
manneohrstrom: great!

python/task_manager/background_task_manager.py
robblau: In a chat with JF today we were talking about doing the select in a subthread and then scheduling the task to run on the right thread without signaling.  To avoid having to poll. We'll see if it is doable.

manneohrstrom: Sounds great if we can avoid polling -- as long as we are conscious of the fact that there may be issues around starting up / shutting down threads at a high frequency.

PR #34, opened 2016-04-15
For #35801, file save freeze fix

jfboismenu:
The polling was replaced by a consumer thread that dispatches the events to the background manager's thread using QMetaObject.invokeMethod, which has shown to be reliable in the past and not introduce deadlocks between PySide and the GIL.

### DIFF
--- a/python/task_manager/background_task_manager.py
+++ b/python/task_manager/background_task_manager.py
@@ -18,7 +18,7 @@ from sgtk import TankError
 
 from .background_task import BackgroundTask
 from .worker_thread import WorkerThread
-from .results_poller import ResultsPoller
+from .results_poller import ResultsDispatcher
 
 
 class BackgroundTaskManager(QtCore.QObject):
@@ -28,24 +28,6 @@ class BackgroundTaskManager(QtCore.QObject):
 
     The BackgroundTaskManager class itself is reentrant but not thread-safe so its methods should only
     be called from the thread it is created in. Typically this would be the main thread of the application.
-
-    .. note::
-        Signalling between two different threads in PySide is broken in several versions
-        of PySide. There are very subtle race conditions that arise when there is a lot
-        of signalling between two threads. Some of these things have been fixed in later
-        versions of PySide, but most hosts integrate PySide 1.2.2 and lower, which are
-        victim of this race condition.
-
-        The background task manager does a lot on inter-threads communications and
-        therefore can easily fall pray to these deadlocks that exist within PySide.
-
-        Therefore, the background task manager uses a polling system to process results
-        from the worker threads. The worker threads inserts results inside a queue and
-        the background manager's thread will flush the queue at 100 milliseconds intervals.
-
-        Also note that because there is a 100 milliseconds interval between two polls,
-        it means that the will automatically be a delay between the end of a task
-        and the beginning of another task.
 
     :signal task_completed(uid, group, result): Emitted when a task has been completed.
         The ``uid`` parameter holds the unique id associated with the task,
@@ -104,11 +86,11 @@ class BackgroundTaskManager(QtCore.QObject):
         self._upstream_task_map = {}
         self._downstream_task_map = {}
 
-        # Create the task result poller.
-        self._results_poller = ResultsPoller(self)
-        self._results_poller.task_completed.connect(self._on_worker_thread_task_completed)
-        self._results_poller.task_failed.connect(self._on_worker_thread_task_failed)
-        self._results_poller.start()
+        # Create the results dispatcher
+        self._results_dispatcher = ResultsDispatcher(self)
+        self._results_dispatcher.task_completed.connect(self._on_worker_thread_task_completed)
+        self._results_dispatcher.task_failed.connect(self._on_worker_thread_task_failed)
+        self._results_dispatcher.start()
 
     def next_group_id(self):
         """
@@ -161,8 +143,8 @@ class BackgroundTaskManager(QtCore.QObject):
         self._available_threads = []
         self._all_threads = []
 
-        # Shut down the poller thread
-        self._results_poller.shut_down()
+        # Shut down the dispatcher thread
+        self._results_dispatcher.shut_down()
         self._log("Shut down successfully!")
 
     def add_task(self, cbl, priority=None, group=None, upstream_task_ids=None, task_args=None, task_kwargs=None):
@@ -355,7 +337,7 @@ class BackgroundTaskManager(QtCore.QObject):
         # create a new worker thread - note, there are two different implementations of the WorkerThread class
         # that use two different recipes.  Although WorkerThreadB is arguably more correct it has some issues
         # in PyQt so WorkerThread is currently preferred - see the notes in the class above for further details
-        thread = WorkerThread(self._results_poller, self)
+        thread = WorkerThread(self._results_dispatcher, self)
         if not isinstance(thread, WorkerThread):
             # for some reason (probably memory corruption somewhere else) I've occasionally seen the above
             # creation of a worker thread return another arbitrary object!  Added this in here so the code
--- a/python/task_manager/results_poller.py
+++ b/python/task_manager/results_poller.py
@@ -17,58 +17,42 @@ from sgtk.platform.qt import QtCore
 import sgtk
 
 
-class TaskCompletedEvent(object):
+class _TaskCompletedEvent(object):
     """
     Event sent when a task is succesfully completed.
     """
 
-    def __init__(self, worker_thread, task, result, results_queue):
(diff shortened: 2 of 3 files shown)


python/task_manager/background_task_manager.py
```diff
@@ -18,7 +18,7 @@
 
 from .background_task import BackgroundTask
 from .worker_thread import WorkerThread
-from .results_poller import ResultsPoller
+from .results_poller import ResultsDispatcher
```
jfboismenu: I will rename results_poller when I am done addressing review comments. I've come not to trust GitHub to diff the files correctly when a rename happens.

python/task_manager/results_poller.py
```diff
@@ -9,103 +9,179 @@
...
+        Signals the task_failed event on the results queue.
+        """
+        self._results_queue.task_failed.emit(self._worker_thread, self._task, self._msg, self._traceback)
+
+
+class ResultsDispatcher(QtCore.QThread):
+    """
+    Dispatches events synchronously to the thread that owns this object.
+
+    Signalling between two different threads in PySide is broken in several versions
```
jfboismenu: This blurb used to be in the public documentation of the API because the polling had an impact on performance. Now that events are processed as soon a the main thread receives them it means that the dispatcher thread is just an implementation detail and it can be omitted from the documentation. As far as users are concerned, nothing has changed functionally from 4.1.0.

python/task_manager/results_poller.py
```diff
@@ -9,103 +9,179 @@
...
         """
+        while True:
+            # Wait for the next result
+            self._event = self._results.get(block=True)
+            # If we're told that we need to quit, do it.
+            if isinstance(self._event, self._ShutdownHint):
+                self._log("Consumer thread received ShutdownHint.")
+                break
+
+            # In order to keep this loop simple, we will assume that the background
```
jfboismenu: Useful wall of text to explain how everything here is thread-safe and simpler this way.

python/task_manager/background_task_manager.py
```diff
@@ -355,7 +337,7 @@ def _get_worker_thread(self):
         # create a new worker thread - note, there are two different implementations of the WorkerThread class
         # that use two different recipes.  Although WorkerThreadB is arguably more correct it has some issues
```
manneohrstrom: Can this stuff about workerThreadB be deleted? It feels like it refers to some outdated code now.

manneohrstrom: Also, I think we should probably get rid of `WorkerThreadSeparateThread` at this point - do you see any benefit of keeping it at this point?

python/task_manager/results_poller.py
```diff
@@ -9,103 +9,179 @@
...
-        :param worker_thread: Thread that completed the task.
-        :param task: Task that failed.
-        :param msg: Error message from the task,
-        :param traceback_: Call stack from the task.
-        """
-        self._results.put(
-            (worker_thread, task, msg, traceback_)
-        )
+            self._event.execute()
+            self._event
```
manneohrstrom: i think this line should be removed.

python/task_manager/worker_thread.py
```diff
@@ -95,18 +97,19 @@ def run(self):
                     if not self._process_tasks:
                         break
                     # emit the result (non-blocking):
-                    self._results_poller.queue_task_completed(self, task_to_process, result)
+                    self._results_dispatcher.produce_event(TaskCompletedEvent(self, task_to_process, result, self._results_dispatcher))
```
manneohrstrom: this line is very long

python/task_manager/worker_thread.py
```diff
@@ -95,18 +97,19 @@ def run(self):
...
+                print e
                 # something went wrong so emit failed signal:
                 self._mutex.lock()
                 try:
                     if not self._process_tasks:
                         break
                     tb = traceback.format_exc()
                     # emit failed signal (non-blocking):
-                    self._results_poller.queue_task_failed(self, task_to_process, str(e), tb)
+                    self._results_dispatcher.produce_event(TaskFailedEvent(self, task_to_process, str(e), tb, self._results_dispatcher))
```
manneohrstrom: this line is very long

python/task_manager/worker_thread.py
```diff
@@ -95,18 +97,19 @@ def run(self):
                     if not self._process_tasks:
                         break
                     # emit the result (non-blocking):
-                    self._results_poller.queue_task_completed(self, task_to_process, result)
+                    self._results_dispatcher.produce_event(TaskCompletedEvent(self, task_to_process, result, self._results_dispatcher))
                 finally:
                     self._mutex.unlock()
             except Exception, e:
+                print e
```
manneohrstrom: stray debug?

python/task_manager/worker_thread.py
manneohrstrom: Hmm, having looked at the code, it feels to me that we may be making this a lot more complicated just because we are introducing these `TaskCompletedEvent` and `TaskFailedEvent` classes. Do we need them? Can they just be internal to the results dispatcher?

Maybe doing something like this is simpler:

```
self._results_dispatcher.signal_completion(task_to_process, result))
self._results_dispatcher.signal_failure(task_to_process, str(e), tb))
```

This seems conceptually a little clearer to me - maybe we can discuss?

python/task_manager/results_poller.py
```diff
@@ -9,103 +9,179 @@
...
-
-        :param worker_thread: Thread that completed the task.
-        :param task: Task that failed.
-        :param msg: Error message from the task,
-        :param traceback_: Call stack from the task.
-        """
-        self._results.put(
-            (worker_thread, task, msg, traceback_)
-        )
+            self._event.execute()
```
manneohrstrom: Wondering if the generic event classes with their `execute()` method is making the code harder to follow? Could it just be queued up as a simple dictionary instead? Would that make it easier to follow? I think that it's the fact that all `execute()` does is to emit a signal from this class that makes me wonder if we need all the extra object wrapping?

manneohrstrom:
Looks cool! Very exciting that this works and glad the polling is gone 👍 I have a feeling this may be a pattern that we can use in other places. I did have some questions about the structure of the classes, maybe we can discuss in a quick call? Basically wondering if it will be conceptually easier to follow if we pitch this class more like a "emit wrapper" and potentially skip those event classes? Let's chat!

Nice work! 🎏

python/task_manager/background_task_manager.py
```diff
@@ -355,7 +337,7 @@ def _get_worker_thread(self):
...
         # in PyQt so WorkerThread is currently preferred - see the notes in the class above for further details
-        thread = WorkerThread(self._results_poller, self)
+        thread = WorkerThread(self._results_dispatcher, self)
```
manneohrstrom: nitpicking: there are still mentions here about workerthreadB

python/task_manager/worker_thread.py
```diff
@@ -95,7 +94,7 @@ def run(self):
                     if not self._process_tasks:
                         break
                     # emit the result (non-blocking):
-                    self._results_poller.queue_task_completed(self, task_to_process, result)
+                    self._results_dispatcher.emit_completed(self, task_to_process, result)
```
manneohrstrom: nice!

python/task_manager/results_poller.py
```diff
@@ -9,103 +9,182 @@
...
-
-    def queue_task_completed(self, worker_thread, task, result):
+            event = self._event
+            self._event = None
+            if isinstance(event, _TaskCompletedEvent):
+                self.task_completed.emit(event.worker_thread, event.task, event.result)
+            elif isinstance(event, _TaskFailedEvent):
+                self.task_failed.emit(event.worker_thread, event.task, event.message, event.traceback)
+            else:
+                raise Exception("Unknown event type: %s" % type(event).__name__)
```
manneohrstrom: nitpicking: should this raise `ValueError`?

python/task_manager/results_poller.py
```diff
@@ -9,103 +9,182 @@
...
+
+    The background task manager does a lot on inter-threads communications and
+    therefore can easily fall pray to these deadlocks that exist within PySide.
+
+    Therefore, we instead use Qt's QMetaObject invokeMethod to carry information
+    to the background task manager thread in a thread-safe manner, since it
+    doesn't exhibit the bad behaviour from PySide's signals.
+    """
+
+    class _ShutdownHint(object):
```
manneohrstrom: With the latest changes, should this be moved out of the class and be named `_ShutdownRequestedEvent` for consistency with `_TaskFailedEvent` and `_TaskCompletedEvent`?

manneohrstrom:
looks awesome!

PR #37, opened 2016-04-28
Adds support for additional_filter_presets to the ShotgunModel.

thebeeland:
ShotgunModel and SimpleShotgunModel now accept an optional additional_filter_presets argument that is passed down to any API find calls that are run under the hood. If used, it is also incorporated into the caching mechanism's hashing routine.

### DIFF
--- a/python/shotgun_globals/date_time.py
+++ b/python/shotgun_globals/date_time.py
@@ -1,0 +1,49 @@
+# Copyright (c) 2016 Shotgun Software Inc.
+# 
+# CONFIDENTIAL AND PROPRIETARY
+# 
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import datetime
+
+def create_human_readable_timestamp(dt, postfix=""):
+    """
+    Return the time represented by the argument as a string where the date portion is
+    displayed as "Yesterday", "Today", or "Tomorrow" if appropriate.
+
+    By default just the date is displayed, but additional formatting can be appended
+    by using the postfix argument.
+
+    :param dt: The date and time to convert to a string
+    :type dt: :class:`datetime.datetime` or float
+
+    :param postfix: What will be displayed after the date portion of the dt argument
+    :type postfix: A strftime style String
+
+    :returns: A String representing dt appropriate for display
+    """
+    # shotgun_model converts datetimes to floats representing unix time so
+    # handle that as a valid value as well
+    if isinstance(dt, float):
+        dt = datetime.datetime.fromtimestamp(dt)
+
+    # get the delta and components
+    delta = datetime.datetime.now(dt.tzinfo) - dt
+
+    if delta.days == 1:
+        format = "Yesterday%s" % postfix
+    elif delta.days == 0:
+        format = "Today%s" % postfix
+    elif delta.days == -1:
+        format = "Tomorrow%s" % postfix
+    else:
+        # use the date formatting associated with the current locale
+        format = "%%x%s" % postfix
+
+    time_str = dt.strftime(format)
+    return time_str
+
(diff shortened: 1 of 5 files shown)


jfboismenu:
If a client using an older version of the Shotgun API who's `find` doesn't support `additional_filter_presets` is using this code out of the box, they will get a `TypeError` because of an unexpected argument. If they have the right version of the Shotgun API, but their server doesn't support it (aka most local clients), it will [raise an exception](https://github.com/shotgunsoftware/python-api/blob/35229_add_additional_server_presets/shotgun_api3/shotgun.py#L650-L651).

We should support this potential argument inside `SgDataRetriever._task_execute_find` in a backwards compatible manner so that versions of the API that don't even know of the argument will see it removed from the `kwargs` and versions that's do will make sure that the server does support the argument. 

I would not require clients to upgrade core in order to ensure that they have a version of the Shotgun API that supports the argument, as almost every single app requires this framework and clients expecting bugfixes to this framework will not necessarily want to upgrade core.

python/shotgun_globals/date_time.py
```diff
@@ -0,0 +1,49 @@
...
+
+    :param postfix: What will be displayed after the date portion of the dt argument
+    :type postfix: A strftime style String
+
+    :returns: A String representing dt appropriate for display
+    """
+    # shotgun_model converts datetimes to floats representing unix time so
+    # handle that as a valid value as well
+    if isinstance(dt, float):
+        dt = datetime.datetime.fromtimestamp(dt)
```
manneohrstrom: A unix timestamp is defined as the number of seconds that have elapsed since 00:00:00 Coordinated Universal Time (UTC), Thursday, 1 January 1970. So I *think* with this code you will be comparing a unix timestamp that is in the UTC timezone with `datetime.datetime.now` in your local timezone. I think `datetime.datetime.fromtimestamp(dt)` creates a a datetime object with "empty" time zone def.

One way to fix this is to use the shotgun UTC timezone object when creating the unix time datetime object. Then I think the code would work. But I might be wrong about all of the above - just wanted to double check - python time zones are so confusing to me. :)

python/shotgun_model/simple_shotgun_model.py
```diff
@@ -67,10 +70,21 @@ def load_data(self, entity_type, filters=None, fields=None, order=None, limit=No
                   is handling, allowing a user to for example show the twenty most recent notes or
                   similar.
         :param columns: List of Shotgun fields to use to populate the model columns
+        :param additional_filter_presets: List of Shotgun filter presets to apply, e.g.
+                  ``[{"preset_name":"LATEST","latest_by":"BY_PIPELINE_STEP_NUMBER_AND_ENTITIES_CREATED_AT"}]``
```
manneohrstrom: This looks quite magical to me. Are there any type of doc reference or context we can provide with info about what these additional presets mean, where you can find a list of what's available and how they are meant to be used?

python/shotgun_globals/date_time.py
```diff
@@ -0,0 +1,49 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import datetime
+
+def create_human_readable_timestamp(dt, postfix=""):
+    """
+    Return the time represented by the argument as a string where the date portion is
+    displayed as "Yesterday", "Today", or "Tomorrow" if appropriate.
+
+    By default just the date is displayed, but additional formatting can be appended
```
manneohrstrom: a list of example outputs would be useful here.

python/shotgun_globals/date_time.py
```diff
@@ -0,0 +1,49 @@
...
+    by using the postfix argument.
+
```
manneohrstrom: Heads up that the activity stream has got a similar logic:

https://github.com/shotgunsoftware/tk-framework-qtwidgets/blob/aa8ed8c49814dfea96c0cf5863d598de97a42a39/python/activity_stream/widget_activity_stream_base.py#L106

This is trying to make the string representation as simple as possible:

- if the time is more than one year ago, display on the form `26 June 2012`
- if the time is more than one week ago, display on the form `26 June`
- if the time is more than one day ago, display it as day of week - e.g. `Sunday`
- if the time is from earlier today, display it as a timestamp, e.g. `23:22`

Maybe the use case is different for this type of method? Maybe at some point we should add this logic to the shotgun global too?

btw, i noticed that that code has got the same potential UTC unixtime stamp issue, so maybe our unix times are actually not stored in UTC as they ideally should be, but stored in the local tz.... Maybe something to discuss.

PR #39, opened 2016-05-17
For #36192, PySide2 compatibility layer

jfboismenu:
- PySide2 compatibility
- updated out of date documentation

### DIFF
--- a/python/shotgun_model/shotgun_model.py
+++ b/python/shotgun_model/shotgun_model.py
@@ -492,6 +492,12 @@ class ShotgunModel(QtGui.QStandardItemModel):
         params_hash.update(str(self.__fields))
         params_hash.update(str(self.__order))
         params_hash.update(str(self.__hierarchy))
+        # If this value changes over time (like between Qt4 and Qt5), we need to
+        # assume our previous user roles are invalid since Qt might have taken over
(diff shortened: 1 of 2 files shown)


python/shotgun_model/shotgun_model.py
```diff
@@ -492,6 +492,12 @@ def _load_data(self, entity_type, filters, hierarchy, fields, order=None, seed=N
         params_hash.update(str(self.__fields))
         params_hash.update(str(self.__order))
         params_hash.update(str(self.__hierarchy))
+        # If this value changes over time (like between Qt4 and Qt5), we need to
+        # assume our previous user roles are invalid since Qt might have taken over
+        # it. If role's value is 32, don't add it to the hash so we don't
+        # invalidate PySide/PyQt4 caches.
+        if QtCore.Qt.UserRole != 32:
+            params_hash.update(str(QtCore.Qt.UserRole))
```
robblau: Is it worth going a little bigger with the value we use for hashing here?  This assumes that everything other than the different UserRole is compatible between Qt4/Qt5.  Should we just has on the major version of Qt so that when (if) Qt6 comes out we'd give it its own space?

Not critical, but feels like a more appropriate value to use.

PR #40, opened 2016-05-25
#34772 Adds utility methods to the cached schema API

josh-t:
Utility methods added primarily for use with the qtwidget's shotgun_fields.

### DIFF
--- a/python/shotgun_globals/cached_schema.py
+++ b/python/shotgun_globals/cached_schema.py
@@ -456,6 +456,17 @@ class CachedShotgunSchema(QtCore.QObject):
 
     @classmethod
     def get_valid_values(cls, sg_entity_type, field_name):
+        """
+        Returns valid values for fields with a list of choices.
+
+        :param str sg_entity_type: The entity type.
+        :param str field_name: The name of the field on the entity
+
+        :return: A :obj:`list` of valid values defined by the schema
+        :rtype list:
+
+        :raises: ``ValueError`` if the field has no valid values.
+        """
         self = cls.__get_instance()
         self._check_schema_refresh(sg_entity_type, field_name)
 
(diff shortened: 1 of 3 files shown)


python/shotgun_globals/cached_schema.py
```diff
@@ -507,4 +534,43 @@ def get_status_color(cls, status_code):
...
+
+        :returns: ``True`` if the field is ediable, ``False`` otherwise.
+        :rtype: :obj:`bool`
+        """
+        self = cls.__get_instance()
+        self._check_schema_refresh(sg_entity_type, field_name)
+
+        if sg_entity_type in self._type_schema and field_name in self._field_schema[sg_entity_type]:
+            data = self._field_schema[sg_entity_type][field_name]
+            return data.get("editable", {}).get("value", False)
```
robblau: Is there a case where editable isn't there?
I'm just wondering what it is, since the default to False here could have implications.

I think examples would help clarify if this should be False, True, or an error condition.

python/shotgun_globals/cached_schema.py
```diff
@@ -507,4 +534,43 @@ def get_status_color(cls, status_code):
...
+
+        :returns: ``True`` if the field is visible, ``False`` otherwise.
+        :rtype: :obj:`bool`
+        """
+        self = cls.__get_instance()
+        self._check_schema_refresh(sg_entity_type, field_name)
+
+        if sg_entity_type in self._type_schema and field_name in self._field_schema[sg_entity_type]:
+            data = self._field_schema[sg_entity_type][field_name]
+            return data.get("visible", {}).get("value", True)
```
robblau: Same comment as above, when is visible missing?

python/shotgun_globals/cached_schema.py
josh-t: I'm not sure. I'm fine switching this over to error out if not available.

josh-t:
manually merged.

PR #44, opened 2016-06-28
#36882: Adds attachment downloading capabilities to the data retriever.

thebeeland:
Much like we need to download thumbnail representations of various entities, we also need easy, asynchronous access to data associated with an Attachment entity in Shotgun. This follows the thumbnail-downloading pattern that already exists in the data retriever class.

### DIFF
--- a/python/shotgun_data/shotgun_data_retriever.py
+++ b/python/shotgun_data/shotgun_data_retriever.py
@@ -84,10 +84,14 @@ class ShotgunDataRetriever(QtCore.QObject):
     # Note: a higher value means more important and will get run before lower 
     # priority tasks
 
+    # Attachment checks and downloads are more important than thumbnails,
+    # as having access to that data will often be required instead of as
+    # a nice-to-have. As a result, this gets a bit more priority.
+    _CHECK_ATTACHMENT_PRIORITY      = 55
+
     # thumbnail checks are local disk checks and very fast.  These 
     # are always carried out before any shotgun calls
     _CHECK_THUMB_PRIORITY           = 50
-    _CHECK_ATTACHMENT_PRIORITY      = 50
 
     # the shotgun schema is often useful to have as early on as possible,
     # sometimes other shotgun operations also need the shotgun schema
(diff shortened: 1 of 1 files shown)


python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -87,6 +87,7 @@ class ShotgunDataRetriever(QtCore.QObject):
     # thumbnail checks are local disk checks and very fast.  These 
     # are always carried out before any shotgun calls
     _CHECK_THUMB_PRIORITY           = 50
+    _CHECK_ATTACHMENT_PRIORITY      = 50
```
manneohrstrom: Feels like attachment should have slightly higher prio than thumbs. There are fewer attachments than there are thumbs and generally i think in a lot of use cases the attachment is 'required' where a thumbnail is often 'nice to have'.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -101,6 +102,7 @@ class ShotgunDataRetriever(QtCore.QObject):
     # thumbnails are downloaded last as they are considered low-priority 
     # and can take a relatively significant amount of time
     _DOWNLOAD_THUMB_PRIORITY        = 20
+    _DOWNLOAD_ATTACHMENT_PRIORITY   = 20
```
manneohrstrom: same comment as above - attachment should have slightly higher prio than thumbs I think.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -408,10 +411,45 @@ def _add_task(self, task_cb, priority, task_args=None, task_kwargs=None):
...
+        """
+        Downloads an attachment from Shotgun asynchronously or returns a cached
+        file path if found.
+
+        :param dict attachment_entity: The Attachment entity to download data from.
+
+        :returns: A unique identifier representing this request.
+        """
+        if not self._task_manager:
+            return
```
manneohrstrom: Maybe a log warning emitted here? (and in register thumb)

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -408,10 +411,45 @@ def _add_task(self, task_cb, priority, task_args=None, task_kwargs=None):
                                               task_kwargs = task_kwargs)
         return str(task_id)
 
+    def request_attachment(self, attachment_entity):
+        """
+        Downloads an attachment from Shotgun asynchronously or returns a cached
+        file path if found.
+
+        :param dict attachment_entity: The Attachment entity to download data from.
```
manneohrstrom: what dict fields are required here? Looked at the code and looks like it needs a `this_file` key - would be great with some details and an example. Do we need to be concerned with S3 ids timing out?

manneohrstrom:
CR done! Looking 🍰 🍸 ! Back to you Jeff!

PR #45, opened 2016-06-28
#35963 Allows editor widgets to be used as delegates with the shotgunmodel

josh-t:
This changes adds the ability to specify columns in the SG model whose items should have the `ItemIsEditable` flag set. This allows for the item to be edited within views backed by the model.

### DIFF
--- a/python/shotgun_model/shotgun_model.py
+++ b/python/shotgun_model/shotgun_model.py
@@ -384,7 +384,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
 
     def _load_data(
         self, entity_type, filters, hierarchy, fields, order=None, seed=None, limit=None,
-        columns=None, additional_filter_presets=None
+        columns=None, editable_columns=None, additional_filter_presets=None
     ):
         """
         This is the main method to use to configure the model. You basically
@@ -440,6 +440,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
                                           each column in the row contains the value for the corresponding field from columns. This means
                                           that the data from the loaded entity will be available field by field. Subclasses can modify
                                           this behavior by overriding _get_additional_columns.
+        :param editable_columns:          A subset of ``columns`` that will be editable in views that use this model.
         :param additional_filter_presets: List of Shotgun filter presets to apply, e.g.
                                           ``[{"preset_name":"LATEST","latest_by":"BY_PIPELINE_STEP_NUMBER_AND_ENTITIES_CREATED_AT"}]``
 
@@ -458,6 +459,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
         self.__order = order or []
         self.__hierarchy = hierarchy
         self.__column_fields = columns or []
+        self.__editable_fields = editable_columns or []
         self.__limit = limit or 0 # 0 means get all matches
         self.__additional_filter_presets = additional_filter_presets
 
@@ -529,6 +531,7 @@ class ShotgunModel(QtGui.QStandardItemModel):
         self.__log_debug("Fields: %s" % self.__fields)
         self.__log_debug("Order: %s" % self.__order)
         self.__log_debug("Columns: %s" % self.__column_fields)
+        self.__log_debug("Editable Columns: %s" % self.__editable_fields)
         self.__log_debug("Filter Presets: %s" % self.__additional_filter_presets)
 
         self.__log_debug("First population pass: Calling _load_external_data()")
@@ -924,7 +927,8 @@ class ShotgunModel(QtGui.QStandardItemModel):
             data = get_sg_data(primary_item)
             for column in columns:
                 # set the display role to the string representation of the value
-                column_item = QtGui.QStandardItem(self._generate_display_name(column, data))
+                column_item = ShotgunStandardItem(self._generate_display_name(column, data))
+                column_item.setEditable(column in self.__editable_fields)
 
                 # set associated field role to be the column value itself
                 value = data.get(column)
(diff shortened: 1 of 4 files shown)


python/shotgun_model/shotgun_model.py
```diff
@@ -440,6 +440,7 @@ def _load_data(
                                           each column in the row contains the value for the corresponding field from columns. This means
                                           that the data from the loaded entity will be available field by field. Subclasses can modify
                                           this behavior by overriding _get_additional_columns.
+        :param editable_columns:          A subset of ``columns`` that will be editable in views that use this model.
```
robblau: Any reason to restrict it to a subset rather than just having the full column list be the union?

Otherwise there should be a check to make sure it is a subset since I think passing in invalid data will result in a kinda unintuitive error.

python/shotgun_model/shotgun_model.py
```diff
@@ -384,7 +384,7 @@ def reset(self):
 
     def _load_data(
         self, entity_type, filters, hierarchy, fields, order=None, seed=None, limit=None,
-        columns=None, additional_filter_presets=None
+        columns=None, editable_columns=None, additional_filter_presets=None
```
manneohrstrom: what if someone has derived from `ShotgunModel` and subclassed `_load_data` - I think this could potentially break things because you are inserting a new parameter rather than appending it.

python/shotgun_model/shotgun_model.py
manneohrstrom: catching up here but what's the design thinking around putting this in the model vs controlling editability in the view? We often have multiple views hooked up to a single model and there are definitely cases where you want to make one view editable and one not. I think it would be great to just briefly address this use case in the doc string. Also if you could add some type info and maybe a usage example to both the columns and editable_columns param that would be great!

manneohrstrom: Can i make a model which only has got a single column (e.g. like the left hand side loader treeview) but where that column has editable values? Another use case that could be great to cover via a usage example, either as part of the parameter or in the main docstring. A lot of TDs (and developers!) are recipe driven, the more examples we provide, the quicker they will pick up the concepts i think.

python/shotgun_model/simple_shotgun_model.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, parent, bg_task_manager=None):
 
     def load_data(
         self, entity_type, filters=None, fields=None, order=None, limit=None,
-        columns=None, additional_filter_presets=None
+        columns=None, editable_columns=None, additional_filter_presets=None
```
manneohrstrom: same issue here with param order.

python/shotgun_model/simple_shotgun_model.py
```diff
@@ -70,6 +70,7 @@ def load_data(
                   is handling, allowing a user to for example show the twenty most recent notes or
                   similar.
         :param columns: List of Shotgun fields to use to populate the model columns
+        :param editable_columns: A subset of ``columns`` that will be editable in views that use this model.
         :param additional_filter_presets: List of Shotgun filter presets to apply, e.g.
```
manneohrstrom: would be great with some additional info about this additional_filter_presets stuff too - it feel reall mysterious! Is there a reference where i can go look and see what the heck `BY_PIPELINE_STEP_NUMBER_AND_ENTITIES_CREATED_AT` means and what other things are available?

python/shotgun_model/simple_shotgun_model.py
```diff
@@ -70,6 +70,7 @@ def load_data(
                   is handling, allowing a user to for example show the twenty most recent notes or
                   similar.
         :param columns: List of Shotgun fields to use to populate the model columns
+        :param editable_columns: A subset of ``columns`` that will be editable in views that use this model.
```
manneohrstrom: if we add examples to the main `_load_data` method, we could have a sphinx reference from this method back to the main one.

python/shotgun_model/shotgun_model.py
josh-t: I don't think `additional_filter_presets` are public yet right? They're not documented anyway. I've moved the argument to the end to be safe. :)

python/shotgun_model/shotgun_model.py
josh-t: @robblau: The `columns` param allows you to specify the order of the columns to display. It's also already a subset of the `fields` param. My personal preference would be to keep it this way for consistency, but I'm not passionate about it either way. I've added a check to ensure it is a subset for now. 

@manneohrstrom: The items are created by the model and that's typically where the editable state is set for standard item models. I think the views can define edit triggers to allow disallow for a particular view. That seem reasonable?

Regarding the single, editable column, yes you can do that. I feel like there are so many options here, this may better be relayed via the test harness/example app we need to have. Thoughts?

python/shotgun_model/simple_shotgun_model.py
josh-t: changed here too.

python/shotgun_model/simple_shotgun_model.py
josh-t: It's still a mystery to me as well. I think this is all server side now and undocumented. I'll add to Wed meeting agenda for explanation/discussion.

python/shotgun_model/shotgun_model.py
manneohrstrom: The harness is great but there is also value in having some usage examples showing up in the sphinx docs. A lot of people will look [here](http://developer.shotgunsoftware.com/tk-framework-shotgunutils/shotgun_model.html#shotgun_model.ShotgunModel._load_data) and having some cut-past type examples I think would be super useful. Maybe we can mention the test harness? or add a comment to the ticket covering the test harness to make sure we update these docs to include a link to the test harness once it goes public!

josh-t: I've added an example and the test app ticket now suggests we link to that code once we have examples in place.

PR #46, opened 2016-07-15
#37467 Updates field editable/visible to use project id based schema caches

josh-t:
This changes updates the `field_is_editable` and `field_is_visible` methods to use the new style schema caches which key off of the `project_id`.

### DIFF
--- a/python/shotgun_globals/cached_schema.py
+++ b/python/shotgun_globals/cached_schema.py
@@ -103,16 +103,17 @@ class CachedShotgunSchema(QtCore.QObject):
 
     def _get_current_project_id(self):
         """
-        Gets the project id associated with the current context, or 0
-        if operating in a site-level context.
+        Return the id of the current project.
 
-        :returns:   int
+        :returns: The project id associated with the current context, or ``None``
+            if operating in a site-level context.
+        :rtype: ``int`` or ``None``
         """
-        # The project id is going to be passed around by signals, as
-        # well as being passed to a core hook in some situations. As
-        # a result, we need it to always be an int value.
+
         if self._bundle.tank.pipeline_configuration.is_site_configuration():
-            project_id = 0
+            # site configuration (no project id). Return None which is
+            # consistent with core.
+            project_id = None
         else:
             project_id = self._bundle.tank.pipeline_configuration.get_project_id()
 
--- a/python/shotgun_globals/cached_schema.py
+++ b/python/shotgun_globals/cached_schema.py
@@ -744,8 +744,10 @@ class CachedShotgunSchema(QtCore.QObject):
 
         :param str sg_entity_type: the entity type
         :param str field_name: the field name to check editibility
-        :param project_id:  The id of the project.
-                            If None, the current context's project will be used.
+        :param project_id:  The project Entity id. If None, the current
+                            context's project will be used, or the "site"
+                            cache location will be returned if the current
+                            context does not have an associated project.
 
         :returns: ``True`` if the field is ediable, ``False`` otherwise.
         """
@@ -777,8 +779,10 @@ class CachedShotgunSchema(QtCore.QObject):
 
         :param sg_entity_type: the entity type
         :param field_name: the field name to check visibility
-        :param project_id:  The id of the project.
-                            If None, the current context's project will be used.
+        :param project_id:  The project Entity id. If None, the current
+                            context's project will be used, or the "site"
+                            cache location will be returned if the current
+                            context does not have an associated project.
 
         :returns: ``True`` if the field is visible, ``False`` otherwise.
         """
--- a/python/shotgun_globals/cached_schema.py
+++ b/python/shotgun_globals/cached_schema.py
@@ -738,43 +738,63 @@ class CachedShotgunSchema(QtCore.QObject):
         return status_color
 
     @classmethod
-    def field_is_editable(cls, sg_entity_type, field_name):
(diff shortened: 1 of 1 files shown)


python/shotgun_globals/cached_schema.py
```diff
@@ -738,43 +738,63 @@ def get_status_color(cls, status_code, project_id=None):
...
+                            If None, the current context's project will be used.
 
         :returns: ``True`` if the field is ediable, ``False`` otherwise.
         """
         self = cls.__get_instance()
-        self._check_schema_refresh(sg_entity_type, field_name)
 
-        if sg_entity_type in self._type_schema and field_name in self._field_schema[sg_entity_type]:
-            data = self._field_schema[sg_entity_type][field_name]
+        project_id = project_id or self._get_current_project_id()
```
manneohrstrom: this code looks repeated in (at least?) two places. Maybe it can be some sort of helper method?

python/shotgun_globals/cached_schema.py
```diff
@@ -738,43 +738,63 @@ def get_status_color(cls, status_code, project_id=None):
...
     @classmethod
-    def field_is_editable(cls, sg_entity_type, field_name):
+    def field_is_editable(cls, sg_entity_type, field_name, project_id=None):
         """
         Returns a boolean identifying the editability of the entity's field.
 
         :param str sg_entity_type: the entity type
         :param str field_name: the field name to check editibility
+        :param project_id:  The id of the project.
+                            If None, the current context's project will be used.
```
manneohrstrom: hmm, wondering if this is the right logic? Historically, the logic was such that there was a global schema that worked everywhere. Then we added per-project support as an optional setting - see the docs for the shotgun API:

https://github.com/shotgunsoftware/python-api/blob/master/shotgun_api3/shotgun.py#L1213

I think it probably makes more sense to be consistent with the underlying Shotgun API that we are caching here and interpret `project_id=None` as "give me the site-wide setting" rather than "give me the current project setting". Calling code can easily determine the context for the current project.

Generally, we are moving in a direction where we will be working increasingly in a site context (because of TaaP) and should be mindful about the case where we don't have a current project.

python/shotgun_globals/cached_schema.py
manneohrstrom: Heads up I noticed that`self._get_current_project_id()` returns 0 for the site context (which is confusing, i would expect `None`!) - is this something we want to adjust?

python/shotgun_globals/cached_schema.py
```diff
@@ -738,43 +738,63 @@ def get_status_color(cls, status_code, project_id=None):
...
     @classmethod
-    def field_is_visible(cls, sg_entity_type, field_name):
+    def field_is_visible(cls, sg_entity_type, field_name, project_id=None):
         """
         Returns a boolean identifying the visibility of the entity's field.
 
         :param sg_entity_type: the entity type
         :param field_name: the field name to check visibility
+        :param project_id:  The id of the project.
+                            If None, the current context's project will be used.
```
manneohrstrom: same here about confusing logic around site vs. project.

python/shotgun_globals/cached_schema.py
manneohrstrom: Looked at existing methods in the class and they say

```
        :param project_id:  The project Entity id. If None, the current
                            context's project will be used, or the "site"
                            cache location will be returned if the current
                            context does not have an associated project.
```

This seems good to me - maybe all we need to do is to make sure the docstring better reflect 
the actual logic in the case when a site context is being used?

josh-t: Ah! Makes sense. Updated!

python/shotgun_globals/cached_schema.py
josh-t: pinging @thebeeland for his input here.

python/shotgun_globals/cached_schema.py
josh-t: updated!

python/shotgun_globals/cached_schema.py
thebeeland: I don't remember why I chose 0. The implication would be that the cache location would change to the following:

`C:\Users\beelanj\AppData\Roaming\Shotgun\jeff-beeland\p0.dynamite\fw-shotgunutils`

Otherwise I don't think there would be any visible change if we used `None` for the site level.

thebeeland: Actually, the path wouldn't even change, due to this:

https://github.com/shotgunsoftware/tk-core/blob/master/hooks/cache_location.py#L152-L153

manneohrstrom: ah perfect! I think that logic in core has been changing and changing and changing and has finally settled with the release. Perfect, so i suggest switching to None in shotgunutils just for clarity - it feels more natural to me - but it's not a big deal, more an observation...

manneohrstrom:
looks good to me! Added one comment about perhaps considering switching to None.

python/shotgun_globals/cached_schema.py
josh-t: Changed to return `None`.

manneohrstrom: great! heads up that this is what i am seeing right now w caches:

image

I *think* with this none change, the `p0` will go away. which is nice!

Cheers all!

PR #48, opened 2016-07-22
#37604 adjusted debug in task manager

manneohrstrom:
The debug logs are too noisy right now, filled with task information from the task manager.
Dialled down the noise level and added a setting, allowing developers to switch on this debug info if needed. It's useful when debugging concurrency issues but not useful generally.

### DIFF
--- a/python/task_manager/background_task_manager.py
+++ b/python/task_manager/background_task_manager.py
@@ -20,6 +20,11 @@ from .background_task import BackgroundTask
 from .worker_thread import WorkerThread
 from .results_poller import ResultsDispatcher
 
+# Set to True to enable extensive debug logging.
+# Useful for debugging concurrency issues.
+# Do not release with this setting set to True!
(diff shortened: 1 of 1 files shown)


python/task_manager/background_task_manager.py
```diff
@@ -20,6 +20,11 @@
 from .worker_thread import WorkerThread
 from .results_poller import ResultsDispatcher
 
+# Set to True to enable extensive debug logging.
+# Useful for debugging concurrency issues.
+# Do not release with this setting set to True!
+ENABLE_DETAILED_DEBUG = False
```
jfboismenu: I like environment variables for these sort of things. This way you can never commit the flag with True.

manneohrstrom: i have been to environment variable hell and back several times in my carreer. i think if the need ever arises, we can add it. One of Toolkit's original design goals was "no environment variables!" :)

PR #49, opened 2016-08-01
#37113 Shotgun Hierarchy Model

josh-t:
screen shot 2016-08-01 at 5 04 39 pm

These changes introduce a new `ShotgunHierarchyModel` which retrieves data via the new shotgun python-api call `nav_expand()`. The structure and organization of the model is very similar to the existing `ShotgunModel` and in fact they now share a new base class called `ShotgunQueryModel`. 

Example `nav_expand()` call:
```
sg.nav_expand(
    "/Project/65",                      # path to item in SG hierarchy (new concept)
    "Version.entity",                   # the entity field you're building a hierarchy to find
    {'Shot': ['created_at', 'code']}),  # as items are returned in hierarchy, include these fields for specified entity types
)
```
Results:
```
{
  "children": [
    {
      "has_children": true, 
      "label": "Assets", 
      "ref": {
        "kind": "entity_type", 
        "value": "Asset"
      }, 
      "target_entities": {
        "additional_filter_presets": [
          {
            "path": "/Project/65/Asset", 
            "preset_name": "NAV_ENTRIES", 
            "seed": {
              "field": "entity", 
              "type": "Version"
            }
          }
        ], 
        "type": "Version"
      }, 
      "url": "/Project/65/Asset"
    }, 
    {
      "has_children": true, 
      "label": "Shots", 
      "ref": {
        "kind": "entity_type", 
        "value": "Shot"
      }, 
      "target_entities": {
        "additional_filter_presets": [
          {
            "path": "/Project/65/Shot", 
            "preset_name": "NAV_ENTRIES", 
            "seed": {
              "field": "entity", 
              "type": "Version"
            }
          }
        ], 
        "type": "Version"
      }, 
      "url": "/Project/65/Shot"
    }
  ], 
  "has_children": true, 
  "label": "Big Buck Bunny", 
  "parent_url": "/", 
  "ref": {
    "kind": "entity", 
    "value": {
      "id": 65, 
      "type": "Project"
    }
  }, 
  "target_entities": {
    "additional_filter_presets": [
      {
        "path": "/Project/65", 
        "preset_name": "NAV_ENTRIES", 
        "seed": {
          "field": "entity", 
          "type": "Version"
        }
      }
    ], 
    "type": "Version"
  }, 
  "url": "/Project/65"
}
```
Basically, the results give you information about a single item including its `url`, `label`, `parent_url`, and a `ref` dictionary that tells you what the item represents. The current `kind`s are:

- `entity`: A concrete entity instance
- `entity_type`: A container for the type of entity (ex: "Asset", "Shot", etc)
- `list`: A container for other items 
- `no_entity`: A container for items with no parent entity (ex: "Shots with no Sequence")
- `None`: A placeholder that represents no items (ex: "No Shots")

In addition, the returned data has a `has_children` key that tells you if the item has children, and the `children` key gives you a list of immediate children with the same information (minus their list of children). 

From this data, the new model implements the `QAbstractItemModel`'s `hasChildren`, `canFetchMore` and `fetchMore` methods to build a lazy loaded hierarchy for the initial supplied path. 

Each item in the `ShotgunQueryModel` is tracked via a unique id. For the `ShotgunModel`, this is the `id` field of the entities. For the `ShotgunHierarchyModel` this is the `url` field returned via the query. 

Additional changes:
- Added a `SimpleShotgunModel` class that inherits `ShotgunHierarchyModel` that allows for very simple creation of a hierarchy model. Example:
```
    nav_model = shotgun_model.SimpleShotgunHierarchyModel(self)
    nav_model.load_data("/Project/65", "Version.entity")
```
- The `ShotgunDataRetriever` has a new method called `execute_nav_expand()` which corresponds to the new python api call. 
- There is a new utility method on `shotugnutils` called `color_mix`. This method allows for weighted combining of `QColor`s. For this particular PR, the method is used to blend the palette foreground and background colors to create a "greyed out" appearance for some text.

There is still some work to do on the `tk-framework-qtwidgets` side to finish a simple widget that uses this model, but this has probably been too long without fresh eyes. Things to consider for CR:

- The `ShotgunModel` now has a base class, so it's possible that I did something screwy. I took a lot of time to make sure the behavior and API were unchanged, but its possible I missed something.
- The caching mechanism is now a part of the base class and both the hierarchy and regular model share the logic (with the exception of the logic to build the path). This seemed like the best short term solution, but we have been discussing sqlite3 as a possible alternative. 
- Because the existing logic wrote the cache to disk each time the model changed, the hierarchy model is cached each time an item is expanded for the first time. This does not seem to be a bottle neck, but perhaps we should take an alternate approach?
- Some of the private concepts in the `ShotgunModel` have been exposed as protected methods in order to make the base class work. Should make sure the changes make sense from an API standpoint
- The nav API uses the term `path` in the docs but returns `url` fields in the return data. Consider how we should present this API. Perhaps we should push back for consistency on the Platform side since this is still unreleased.


manneohrstrom:
after reading the description, all i can say is: this looks total badass!

manneohrstrom:
nitpicky question: should `python/shotgun_globals/resources/icon_None.png` be called `python/shotgun_globals/resources/icon_None_dark.png` for consistency?

python/shotgun_model/shotgun_model_errors.py
```diff
@@ -0,0 +1,22 @@
...
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+
+
+class ShotgunModelError(sgtk.TankError):
```
manneohrstrom: should we call it `errors.py`. That's consistent with other places.

python/util.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# CONFIDENTIAL AND PROPRIETARY
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from sgtk.platform.qt import QtGui
```
manneohrstrom: The location of this is not right I think. Maybe we can move it into `shotgun_model/utils.py`? Or we should make a new module (folder) for utils.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,979 @@
...
+
+    # ---- caching/serialization related costants
+
+    # magic number for IO streams
+    _FILE_MAGIC_NUMBER = 0xDEADBEEF
+
+    # version of binary format
+    _FILE_VERSION = 22
+
+    def __init__(self, parent, bg_task_manager=0):
```
manneohrstrom: what's the deal with `=0`? should it be `None`? Also needs a docstring.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,979 @@
...
+    # protected properties
+
+    @property
+    def _cache_path(self):
+        """
+        Returns the cache path on disk for this instance.
+        """
+        return self._full_cache_path
+
+    @_cache_path.setter
```
manneohrstrom: py25!

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,979 @@
...
+            return False
+
+        logger.debug(
+            "Now attempting cached data load from: %s ..." %
+            (self._cache_path,)
+        )
+
+        try:
+            time_before = time.time()
+            num_items = self._load_from_disk()
```
manneohrstrom: it seems `load_from_disk` isn't called from any subclasses so I think it should be made private, e.g. `__load_from_disk`. It's confusing, normally we handle "private" methods with underscores but in the case of class inheritance, we should try to make as much as possible `__private` and lave the `_single_underscore` only for methods intended to be subclassed (and official part of the supported interface).

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,979 @@
...
+
+        This method currently handles:
+
+            - Converting datetime objects to universal time stamps.
+
+        :param sg_data:
+        :return:
+        """
+
+        # QT is struggling to handle the special timezone class that the
```
manneohrstrom: this issue has now been fixed in the shotgun API - wonder if we should bite the bullet and clean this up? Or maybe a separate ticket?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,847 @@
...
+        return child_item_count == 0 and item_data.get("has_children", False)
+
+    ############################################################################
+    # protected methods
+
+    def _create_item(self, data, parent=None, row=None):
+        """
+        Creates a model item given the supplied data and optional parent.
+
+        The supplied ``data`` corresponds to the results of a call to the
```
manneohrstrom: For every method that is protected and meant to be subclassed, I suggest we provide a small story in the docstring, explaining when it would be useful to subclass this method. I tried to do that with the Shotgun model and I think it helps both maintain and explore the code.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,847 @@
...
+
+        if row is not None:
+            parent.insertRow(row, item)
+        else:
+            # example of using sort/filter proxy model
+            parent.appendRow(item)
+
+        return item
+
+    def _get_data_cache_path(self, cache_seed=None):
```
manneohrstrom: should probably be private?

python/shotgun_model/shotgun_model.py
```diff
@@ -479,8 +319,47 @@ def _load_data(
...
+        # set our headers
+        headers = [self.FIRST_COLUMN_HEADER] + self._get_additional_column_headers(
+            self.__entity_type,
+            self.__column_fields,
+        )
+        self.setHorizontalHeaderLabels(headers)
+
+        return self._load_cached_data()
+
+    def _get_data_cache_path(self, cache_seed=None):
```
manneohrstrom: should probably be private?

josh-t:
icon is now `python/shotgun_globals/resources/icon_None_dark.png`. :)

python/shotgun_model/shotgun_model_errors.py
josh-t: fixed.

python/util.py
josh-t: added a new `utils` module since the method is useful in general (I think).

python/shotgun_model/shotgun_query_model.py
josh-t: I think this is leftover, cut/paste error from `schema_generation=0` arg. Fixed and updated docstring.

python/shotgun_model/shotgun_query_model.py
josh-t: oops! fixed.

python/shotgun_model/shotgun_query_model.py
josh-t: fixed this method. doing a pass to make sure everything else looks reasonable.

python/shotgun_model/shotgun_query_model.py
josh-t: I'll just remove it unless you object.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: yep. made this one private! :P

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: yep. done.

python/shotgun_model/shotgun_model.py
josh-t: yep. done.

python/shotgun_model/shotgun_query_model.py
manneohrstrom: yes pls remove it! I think we should probably do the same for the shotgun utils - could you take a look and potentially create a ticket?

python/utils/color.py
```diff
@@ -0,0 +1,40 @@
...
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from sgtk.platform.qt import QtGui
+
+
+def color_mix(c1, c1_strength, c2, c2_strength):
```
manneohrstrom: by putting this in the top level of shotgun utils i think we are making it a "component" that people can import into their apps - so we should make sure that when this is imported by for example the hierarchy model, it also uses import_module() rather than a .. style import (for the sphinx build).

For the shotgun model, i wasn't sure if my utility methods would be useful outside of the shotgun model so i added it as a `util.py` file here : https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/master/python/shotgun_model/util.py

I think in this case, we should probably do the same and keep it internal within the hierarchy model for the time being. What do you think?

python/shotgun_model/shotgun_hierarchy_item.py
```diff
@@ -0,0 +1,108 @@
...
+from .shotgun_standard_item import ShotgunStandardItem
+
+
+class ShotgunHierarchyItem(ShotgunStandardItem):
+    """
+    A subclass of ``ShotgunStandardItem`` with access to data provided via
+    the ``nav_expand()`` python API calls.
+    """
+
+    def target_entities(self):
```
manneohrstrom: please add an init for good measure.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+        base_color = QtGui.QApplication.instance().palette().base().color()
+        text_color = QtGui.QApplication.instance().palette().text().color()
+        self._empty_item_color = color_mix(text_color, 1, base_color, 2)
+
+    ############################################################################
+    # public methods
+
+    def clear(self):
+        """
+        Removes all items (including header items) from the model and
```
manneohrstrom: I think it would be great if we could point out which of these methods are subclassing QT4 methods and which ones are our own!

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+        This base class implementation handles storing item lookups for
+        efficiency as well as to prevent issues with garbage collection.
+
+        :param item: The item that was just created.
+        :type item: :class:`~PySide.QtGui.QStandardItem`
+        """
+
+        # keep a reference to this object to make GC happy
+        # (pyside may crash otherwise)
+        self._all_tree_items.append(item)
```
manneohrstrom: should this be private? e.g. `self.__all_tree_items`

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+            raise ShotgunModelError(
+                "ShotgunQueryModel subclass does not define the instance attr: "
+                "`SG_DATA_UNIQUE_ID_FIELD`"
+            )
+
+        # intialize the Qt base class
+        super(ShotgunQueryModel, self).__init__(parent)
+
+        # keep a handle to the current app/engine/fw bundle for convenience
+        self._bundle = sgtk.platform.current_bundle()
```
manneohrstrom: given the subclassing going on here, i think we should be very explicit about encapuslation - e.g. all private member variables having the double underscore - single underscore denoting that we can access that variable in deriving classes. Does that make sense?

So the default should be `self.__double_underscore` and for members that we want to expose to deriving classes, we do `self._single_underscore` (but possibly better to do it via properties?)

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+        # importing these locally to not trip sphinx's imports
+        # shotgun_globals is often used for accessing cached schema information
+        # such as entity type and field display values.
+        self._shotgun_globals = self._bundle.import_module("shotgun_globals")
+        self._shotgun_data = self._bundle.import_module("shotgun_data")
+
+        # keep various references to all items that the model holds.
+        # some of these data structures are to keep the GC
+        # happy, others to hold alternative access methods to the data.
+        self._all_tree_items = []
```
manneohrstrom: these guys feel particularly private to me :)

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+        :param data: Dictionary of the SG data associated with the model.
+        """
+        # the default implementation does not set a tooltip
+        pass
+
+    ############################################################################
+    # protected convenience methods. these methods can be used by subclasses
+    # to manipulate and manage data returned from Shotgun.
+
+    def _do_depth_first_tree_deletion(self, node):
```
manneohrstrom: should this be private?

Also, double check the sphinx docs for this one. there are some explicit exclusions in the docs.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+            os.chmod(filename, 0666)
+
+        except Exception, e:
+            logger.warning(
+                "Could not write cache file '%s' to disk: %s" % (filename, e))
+
+        finally:
+            os.umask(old_umask)
+
+    def _save_to_disk_r(self, stream, item, depth):
```
manneohrstrom: please make private.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+        """
+        Recursive tree writer
```
manneohrstrom: pls add docstring

python/shotgun_model/shotgun_standard_item.py
```diff
@@ -8,16 +8,15 @@
...
 from tank.platform.qt import QtGui
 
+from .util import sanitize_qt
+from .util import get_sg_data as util_get_sg_data
 
 class ShotgunStandardItem(QtGui.QStandardItem):
     """
     Special implementation of StandardItem which bridges PyQt and PySide.
-
-    Do not construct this object directly - instead use the ShotgunModel.create_item() method.
```
manneohrstrom: what has changed that makes this ok?

manneohrstrom: i *think* the reason for this comment is because `create_item` will add the item to the internal data structures that keep references to everything to prevent memory leaks, double frees etc. Otherwise you can create a crash by creating one of these items directly then plug it into a model and there will be issues with GC. But I am not 100% sure this is the case. It may be worth adding some further comments here to provide more details. At least in earlier versions of pyside, we found that the MVC stuff was very brittle in subtle ways, hence all these methods and helpers..

python/shotgun_model/simple_shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,64 @@
...
+    remaining items will be queried asynchronously as items are expanded.
+
+    Subsequently call :meth:`load_data` whenever you wish to change the
+    ``nav_expand()`` query associated with the model.
+
+    This class derives from :class:`ShotgunHierarchyModel` so all the
+    customization methods available in the normal :class:`ShotgunHierarchyModel`
+    can also be subclassed from this class.
+    """
+
```
manneohrstrom: i think we should have a constructor for good measure.

manneohrstrom:
General comment: I did some dumb things in the sphinx docs (this was the first thing that got sphinx docs!) - we should back out of that and make sure it's done more properly. Old sins are here: 

https://github.com/shotgunsoftware/tk-framework-shotgunutils/commit/ddc91982031d3e2fdb48c2adb994770eaa94f2a4#diff-45661b7d5083c77690a302e7895eb88dR217

manneohrstrom:
I have a feeling that @josh-t knows more about sphinx than I do, so not sure if this is useful - but for the engine base class i ended up doing this which i thought worked out really well as a way to explicitly curate which protected methods to include in the docs:

https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/platform/engine.py#L74

manneohrstrom:
We should probably do a separate code/doc review of the docs once they are baked out - this is such a big change!

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -216,8 +217,17 @@ def stop(self):
...
             # and disconnect from it:
             self._task_manager.stop_task_group(self._bg_tasks_group)
-            self._task_manager.task_completed.disconnect(self._on_task_completed)
-            self._task_manager.task_failed.disconnect(self._on_task_failed)
+            # make sure we don't get exceptions trying to disconnect if the
+            # signals were never connected.
+            try:
+                self._task_manager.task_completed.disconnect(self._on_task_completed)
+            except (TypeError, RuntimeError):  # was never connected
+                pass
```
manneohrstrom: debug or warning error output here?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -396,6 +406,29 @@ def execute_method(self, method, *args, **kwargs):
                               priority = ShotgunDataRetriever._SG_CALL_PRIORITY,
                               task_kwargs = task_kwargs)
 
+    def execute_nav_expand(self, *args, **kwargs):
+        """
+        Executes a Shotgun ``nav_expand`` query asynchronously.
```
manneohrstrom: a link to the sg api would be great here.

manneohrstrom:
given that this uses new methods in the shotgun API, should be update the manifest constraints?

python/shotgun_globals/cached_schema.py
```diff
@@ -410,9 +410,18 @@ def unregister_bg_task_manager(cls, task_manager):
...
-                data_retriever.work_completed.disconnect(self._on_worker_signal)
-                data_retriever.work_failure.disconnect(self._on_worker_failure)
-                
+
+                # make sure we don't get exceptions trying to disconnect if the
+                # signals were never connected.
+                try:
+                    data_retriever.work_completed.disconnect(self._on_worker_signal)
+                except (TypeError, RuntimeError):  # was never connected
+                    pass
```
manneohrstrom: what's the use case here? May be worth explaining in comments why we can end up here.

python/shotgun_globals/cached_schema.py
```diff
@@ -735,43 +746,63 @@ def get_status_color(cls, status_code, project_id=None):
...
     @classmethod
-    def field_is_visible(cls, sg_entity_type, field_name):
+    def field_is_visible(cls, sg_entity_type, field_name, project_id=None):
         """
         Returns a boolean identifying the visibility of the entity's field.
 
         :param sg_entity_type: the entity type
         :param field_name: the field name to check visibility
+        :param project_id:  The id of the project.
+                            If None, the current context's project will be used.
```
manneohrstrom: does this need merging with Jeff's changes? See https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/master/python/shotgun_globals/cached_schema.py#L774

python/shotgun_model/shotgun_hierarchy_item.py
```diff
@@ -0,0 +1,108 @@
...
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sgtk
+
+from .shotgun_standard_item import ShotgunStandardItem
+
+
+class ShotgunHierarchyItem(ShotgunStandardItem):
+    """
+    A subclass of ``ShotgunStandardItem`` with access to data provided via
```
manneohrstrom: what's the deal construction? Can this be constructed by any process or only by factory methods? I think we should control creation so we can control memory allocation and release. Items inserted via 

https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/49/files#diff-d442a3297316abbae14a00adf3457696R548

Should probably be created via factory of some sort.

python/shotgun_model/shotgun_hierarchy_item.py
```diff
@@ -0,0 +1,108 @@
...
+
+        data = self.data()
+        if not data or self.__kind() == "empty":
+            return None
+
+        return data.get("url")
+
+    # --------------------------------------------------------------------------
+
+    def __kind(self):
```
manneohrstrom: i prefer `__get_kind()` - just `__kind` sounds like a property.

python/shotgun_model/shotgun_hierarchy_item.py
```diff
@@ -0,0 +1,108 @@
...
+        * no_entity: A container for items with no parent entity (ex: "Shots
+            with no Sequence")
+        * empty: A placeholder that represents no items (ex: "No Shots")
+        """
+
+        data = self.data()
+        if not data:
+            return "empty"
+
+        return data.get("ref", {}).get("kind", "empty")
```
manneohrstrom: should this be `return (data.get("ref") or {}).get("kind", "empty")`?

manneohrstrom: is it safe that we essentially threat "empty" and "undefined" the same here? Should we warn if we get blanks back from SG? I think it's worth adding some comments around the situation when we get Null values back from Shotugn here.

python/shotgun_model/shotgun_hierarchy_item.py
```diff
@@ -0,0 +1,108 @@
...
+            },
+        """
+
+        data = self.data()
+        if not data or self.__kind() == "empty":
+            return None
+
+        return data.get("target_entities")
+
+    def url(self):
```
manneohrstrom: This is confusing. The SG APi seems to talk about `path` and it looks like we are rebranding it to be `url` instead. Why? Path feels clerarer to me, but i may be missing parts of the picture here!

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+    def hard_refresh(self):
+        """
+        Clears any caches on disk, then refreshes the data.
+        """
+        if not self._has_query:
+            # no query in this model yet
+            return
+
+        # when data arrives, force full rebuild
+        self._request_full_refresh = True
```
manneohrstrom: can we clear it right now instead of at the next data arrival?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+        # application instance's base and text colors. This will typically
+        # result in a dimmed appearance for these special items indicating that
+        # they are not clickable. This makes goes outside the typical bounds of
+        # the model by pulling the palette colors from the app instance. This
+        # can be overridden in subclasses via ``_finalize_item()`` though.
+        base_color = QtGui.QApplication.instance().palette().base().color()
+        text_color = QtGui.QApplication.instance().palette().text().color()
+        self._empty_item_color = color_mix(text_color, 1, base_color, 2)
+
+    ############################################################################
```
manneohrstrom: please add suitable `__repr__` to all classes. This is super helpful when doing debugging and logging.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+        child_item_count = item.rowCount()
+
+        # we can fetch more if there are no children already and the item
+        # has children.
+        return child_item_count == 0 and item_data.get("has_children", False)
+
+    ############################################################################
+    # protected methods
+
+    def _get_queried_urls_r(self, item):
```
manneohrstrom: what's the use case for subclassing this? Either provide a subclassing example in the docstring or make it private.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+                continue
+
+            urls.append(data["url"])
+
+            # recurse
+            urls.extend(self._get_queried_urls_r(child_item))
+
+        return urls
+
+    def _insert_subtree(self, nav_data):
```
manneohrstrom: looks private to me.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+        :param dict nav_data: A dictionary of item data as returned via async
+            call to ``nav_expand``.
+        """
+
+        item = self.__create_item(nav_data)
+        self._update_subtree(item, nav_data)
+
+    def _item_created(self, item):
+        """
+        Checks to see if the item is an "empty" item and sets the foreground
```
manneohrstrom: This doc string looks a little strange to me - from a user's doc point of view, i think is quite confusing. Base class docs here:

https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/49/files#diff-d442a3297316abbae14a00adf3457696R504

I think it should at least refer to those docs somehow? Sometimes giving feedback around this sort of thing ie easier once you see the built sphinx docs.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+
+        :param item: The :class:`~PySide.QtGui.QStandardItem` that was created.
+        """
+
+        # call base class implementation as per docs
+        super(ShotgunHierarchyModel, self)._item_created(item)
+
+        data = get_sg_data(item)
+
+        if data.get("ref", {}).get("kind") == "empty":
```
manneohrstrom: all these hard coded strings look a bit iffy. Perhaps there could be some class methods on `ShotgunHierarchyItem` which would provide constants or lookups for this stuff?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+        by the ``nav_expand()`` api call.
+
+        :param item: The :class:`~PySide.QtGui.QStandardItem` item to set the
+            icon for.
+        """
+
+        data = get_sg_data(item)
+
+        item_ref = data.get("ref", {})
+        item_kind = item_ref.get("kind")
```
manneohrstrom: `constants.SG_HIERARCHY_KIND_FIELD`?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+
+        if item_kind == "entity":
+            entity_type = item_ref.get("value", {}).get("type")
```
manneohrstrom: What if `item_ref["value"]` is `None`? does that never happen?
Maybe `(item_ref.get("value") or {}).get("type")` instead?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+        elif item_kind == "no_entity":
+            # this is typically items like "Shots with no Sequence"
+            icon = self._folder_icon
+        else:
+            icon = self._none_icon
+
+        if icon:
+            item.setIcon(icon)
+
+    def _query_hierarchy(self, path, seed_entity_field, entity_fields):
```
manneohrstrom: private? or add example to doc string of when it would be useful to subclass it.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+
+        # data differs. set the new data
+        item.setData(sanitize_for_qt_model(new_item_data), self.SG_DATA_ROLE)
+
+        # ensure the label is updated
+        item.setText(data["label"])
+
+        return True
+
+    def _update_subtree(self, item, nav_data):
```
manneohrstrom: private? or add example to doc string of when it would be useful to subclass it.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -0,0 +1,855 @@
...
+        params_hash.update(str(cache_seed))
+
+        # iterate through the sorted entity fields to ensure consistent order
+        for (entity_type, fields) in sorted(self._entity_fields.iteritems()):
+            for field in fields:
+                entity_field_hash.update("%s.%s" % (entity_type, field))
+
+        # convert the seed entity field into a path segment.
+        # example: Version.entity => Version/entity
+        seed_entity_field_path = os.path.join(
```
manneohrstrom: could parts of this method be avaiable on the base class level? Seems like there is some overlap/repetition between the hierarchy  model and the sg model?

manneohrstrom:
Phew! CR done! This looks great, very clean, very well organized! There are a number of comments simply due to the sheer amount of code changes. Overall it looks fantastic! In summary:

- It's confusing that the sg api calls it path and here it's called url. path seems much better to me - but more importantly, if possible it should be consistent.
- it feels like there are still too many protected methods and members that should be private.
- need a sphinx doc review once the CR has settled
- add some constants to avoid repetion of string fields in the hierarchy data set

thanks!

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+
+        elif a != b:
+            return False
+
+        return True
+
+    ############################################################################
+    # additional method used during de/serialization of model contents
+
+    def _save_to_disk(self):
```
manneohrstrom: please add `@filesystem.with_cleared_umask` decorator

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+        """
+        Save the model to disk using QDataStream serialization.
+        This all happens on the C++ side and is very fast.
+        """
+
+        filename = self._cache_path
+
+        old_umask = os.umask(0)
```
manneohrstrom: and with `@filesystem.with_cleared_umask` we can drop this stuff.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -0,0 +1,947 @@
...
+        try:
+            # try to create the cache folder with as open permissions as
+            # possible
+            cache_dir = os.path.dirname(filename)
+            if not os.path.exists(cache_dir):
```
manneohrstrom: replace with `filesystem.ensure_folder_exists(cache_dir, 0777)` (requires core 0.18)

python/shotgun_data/shotgun_data_retriever.py
josh-t: added a log warning.

python/shotgun_data/shotgun_data_retriever.py
josh-t: added.

python/shotgun_globals/cached_schema.py
josh-t: done. also added debug warning.

python/shotgun_globals/cached_schema.py
josh-t: Yes, I added this while that work was going on. When I merge this into master, i'll make sure to keep those changes and not these.

python/shotgun_model/shotgun_hierarchy_item.py
josh-t: Curious why we need the `__init__.py`? I'm fine adding it, just curious what the reasoning is. :)

python/shotgun_model/shotgun_hierarchy_item.py
josh-t: Yeah, this is what I was talking about in the PR notes: 

> The nav API uses the term `path` in the docs but returns `url` fields in the return data. Consider how we should present this API. Perhaps we should push back for consistency on the Platform side since this is still unreleased.

I like path as well, so I've just changed everything to use `path` on our end. I've also opened a ticket with platform to make it the same throughout.

python/shotgun_model/shotgun_hierarchy_item.py
josh-t: changed.

python/shotgun_model/shotgun_hierarchy_item.py
josh-t: I think the line is correct. If `ref` isn't there (it should always be), return an empty `dict`. From whatever `dict` is returned, get the `kind`. I've changed it up to add some comments and instead of using the API's `empty`, I'm now using `None`. It cleans up the calling code too. Seems better now. Thanks!

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: done.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: Those are in a separate section here: https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/49/files#diff-36900d64bb6de2682eab6cf5bb285286R149

I can rearrange and indicate separately if you prefer. Let me know.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: I'm reusing/keeping the logic from the `ShotgunModel`. It seems like the intention was to not completely wipe the UI while the data was being retrieved? Maybe that's preferable? Maybe there's some benefit in terms of drawing the UI in this order?

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: this is now `__get_queried_paths_r`! :)

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: privatized like the government!

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: Agreed. fixed!

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: Yep. I addressed this in conjunction with some of the previous comments. I've added some methods to the `ShotgunHierarchyItem` class as you suggest which really cleaned things up.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: addressed.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: Well, I was thinking subclasses could use this to customize their tree by manually querying/adding specific branches of their site... But maybe that's a little premature. I've made it private for now.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: made private.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: They're using the same technique, but I the values, order, and path are all different enough that I'm not sure there's much help a base class could provide. Let me know if you think that's not true.

python/shotgun_model/shotgun_query_model.py
josh-t: I've added the additional underscore for members that should be private like `__all_tree_items` and `__tree_data`.

The docstrings for the base class outline which single underscore instance members are defined such that subclasses can use them. My preference would be not to use properties for these.

python/shotgun_model/shotgun_query_model.py
josh-t: yep. done.

python/shotgun_model/shotgun_query_model.py
josh-t: yep. done.

python/shotgun_model/shotgun_query_model.py
josh-t: Yep. I think at one point this was used in the subclasses and then I managed to move it down here. I've made it private.

fyi, that's why much of this ended up this way. I made things protected in order for them to be used by the subclasses. Then as I unraveled a bit, I was able to move more calls back to the base class and I just didn't make them private again. So this is good stuff.

python/shotgun_model/shotgun_query_model.py
josh-t: done!

python/shotgun_model/shotgun_query_model.py
josh-t: done.

python/shotgun_model/shotgun_query_model.py
josh-t: done.

python/shotgun_model/shotgun_query_model.py
josh-t: done.

python/shotgun_model/shotgun_query_model.py
josh-t: done.

python/shotgun_model/shotgun_standard_item.py
josh-t: So there wasn't actually a `create_item()` method on `ShotgunModel`. The `__create_item()` method exists, but is private. Maybe that's why I removed it? 

You're right though, there should be some comment. I've added a warning here and in `ShotgunHierarchyItem`. Maybe that's sufficient?

python/shotgun_model/shotgun_hierarchy_item.py
josh-t: I've added a warning here about creating manually, similar to the warning in `ShotgunStandardItem`. That sufficient?

python/shotgun_model/simple_shotgun_hierarchy_model.py
josh-t: Again, curious what the benefit is. Happy to add one. :)

python/utils/color.py
josh-t: I would love for this to be a first class citizen. I actually pulled it from the `tk-multi-pythonconsole` and would use this from there if I could. It's great for writing apps that need to display custom colors that fit within the various DCC's color palettes. Let me know what you think!

PR #50, opened 2016-08-09
Checks disk for cached globals data for non-current projects.

thebeeland:
If getting schema or status information for a project other than that of the current context, we will now check on disk for cached data before going to Shotgun for it. Previously, we only looked on disk for the current project's schema cache, so all non-current projects required a Shotgun query.


manneohrstrom:
cr looks good!

PR #51, opened 2016-09-02
#38172 Allows for graceful failure when Hierarchy API is not available.

josh-t:
If you create a `ShotgunHierarchyModel` and the shotgun connection doesn't support the `nav_expand()` query, a single item will be added to the model that says the version of SG must be `v7.0.2` in order to use the model. A warning is also logged.

Adds a method: `shotgun_model.hierarchy_is_supported()` to allow graceful fallback for clients.

### DIFF
--- a/info.yml
+++ b/info.yml
@@ -21,7 +21,7 @@ display_name: "Shotgun Utilities Framework"
 description: "A Collection of QT objects and utilities to simplify app development and Shotgun connectivity."
 
 # Required minimum versions for this item to run
-requires_shotgun_version: "v7.0.2"
+requires_shotgun_version:
 requires_core_version: "v0.14.58"
 
 # the frameworks required to run this app
--- a/python/shotgun_model/shotgun_hierarchy_model.py
+++ b/python/shotgun_model/shotgun_hierarchy_model.py
@@ -395,17 +395,16 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
             an external string via this parameter which will be added to the
             seed.
 
-        :return:
+        :returns: True if cached data was loaded, False if not.
         """
 
         if not self._hierarchy_is_supported:
             self.clear()
             root = self.invisibleRootItem()
-            item = QtGui.QStandardItem(
-                "WARNING: SG version must be v7.0.2 to use the Hierarchy Model")
+            item = QtGui.QStandardItem("WARNING: SG version must be v7.0.2 or higher to use the Hierarchy Model")
             item.setEditable(False)
             root.appendRow([item])
-            return
+            return False
 
         # we are changing the query
         self.query_changed.emit()
--- a/python/shotgun_model/shotgun_hierarchy_model.py
+++ b/python/shotgun_model/shotgun_hierarchy_model.py
@@ -28,7 +28,14 @@ logger = sgtk.platform.get_logger(__name__)
 
 class ShotgunHierarchyModel(ShotgunQueryModel):
     """
-    A Qt Model representing a Shotgun hierarchy query (``nav_expand()``).
+    A Qt Model representing a Shotgun hierarchy.
+
+    .. warning::
+
+        Use of this model requires version Shotgun ``v7.0.2`` or later.
+        Attempts to construct an instance of this model on an older version of
+        Shotgun will result with a single item in the model saying that
+        Hierarchy model isn't supported. A warning will also be logged.
 
     This class implements a standard :class:`~PySide.QtCore.QAbstractItemModel`
     specialized to hold the contents of a particular Shotgun query. It is cached
@@ -77,6 +84,8 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
         """
 
         super(ShotgunHierarchyModel, self).__init__(parent, bg_task_manager)
+
+        self._hierarchy_is_supported = self._check_hierarchy_is_supported()
 
         self._path = None
         self._seed_entity_field = None
@@ -230,6 +239,9 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
         :type index: :class:`~PySide.QtCore.QModelIndex`
         """
 
+        if not self._hierarchy_is_supported:
+            return False
+
         if not index.isValid():
             return False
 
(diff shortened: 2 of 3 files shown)


python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -498,6 +543,9 @@ def _refresh_data(self):
         view related states such as selection.
         """
 
+        if not self._hierarchy_is_supported:
+            return
```
thebeeland: Should be `return None` since you're returning actual data at the end.

https://www.python.org/dev/peps/pep-0008/#programming-recommendations

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -362,6 +398,15 @@ def _load_data(
...
         """
 
+        if not self._hierarchy_is_supported:
+            self.clear()
+            root = self.invisibleRootItem()
+            item = QtGui.QStandardItem(
+                "WARNING: SG version must be v7.0.2 to use the Hierarchy Model")
+            item.setEditable(False)
+            root.appendRow([item])
+            return
```
thebeeland: Same as below, `return None`

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -362,6 +398,15 @@ def _load_data(
         :return:
         """
 
+        if not self._hierarchy_is_supported:
+            self.clear()
+            root = self.invisibleRootItem()
+            item = QtGui.QStandardItem(
+                "WARNING: SG version must be v7.0.2 to use the Hierarchy Model")
```
thebeeland: This should probably read "...must be v7.0.2 or higher..."

josh-t: addressed.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: fixed here.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: this method doesn't return anything.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -255,6 +267,30 @@ def canFetchMore(self, index):
...
+    def _check_hierarchy_is_supported(self):
+        """
+        Checks the current Shotgun connection to make sure it supports
+        hierarchy queries.
+        """
+
+        # get the current shotgun connection
+        current_engine = sgtk.platform.current_engine()
+        sg_connection = current_engine.shotgun
+        server_caps = sg_connection.server_caps
```
manneohrstrom: minor: I am not sure this `server_caps` is strictly a public member of the API. To me, a fundamentally more pythonic approach would be to try to execute the operation, and look for a `ShotgunUnsupportedOnServer` exception and then gracefully fall back. This means all the version logic would reside inside the SG API and the exception would be the generic mechaism the Shotgun API used to report incompatibilities to calling code. Does that make sense?

What does the API return today if you run this against an old site?

info.yml
```diff
@@ -21,7 +21,7 @@ display_name: "Shotgun Utilities Framework"
 description: "A Collection of QT objects and utilities to simplify app development and Shotgun connectivity."
 
 # Required minimum versions for this item to run
-requires_shotgun_version: "v7.0.2"
+requires_shotgun_version:
```
manneohrstrom: this is confusing. Are we limiting all of the framework to require server 7.0.2 and at the same time adding a graceful fallback in case the server isn't 7.0.2?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -255,6 +267,30 @@ def canFetchMore(self, index):
...
+
+        if not hierarchy_is_supported(sg_connection):
+            # oops, SG version is not compatible
+            cur_sg_version = "v%s" % (
+                ".".join(map(str,server_caps.version)))
+            logger.warning(
```
manneohrstrom: this warning you are composing here could be the exception message raised from the SG API...

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -359,9 +395,17 @@ def _load_data(
...
             seed.
 
-        :return:
+        :returns: True if cached data was loaded, False if not.
         """
 
+        if not self._hierarchy_is_supported:
+            self.clear()
+            root = self.invisibleRootItem()
+            item = QtGui.QStandardItem("WARNING: SG version must be v7.0.2 or higher to use the Hierarchy Model")
```
manneohrstrom: Is this too long? What does it look like in the UI? Normal items would be stuff like `abc_123`, `lighting` etc. right?

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: Uh oh, I've used `server_caps` a few times in different places. :| 

Yeah, we should definitely discuss...

```python
>>> shotgun.server_caps.version
(7, 0, 0)
>>> shotgun.nav_expand("/", "Version.entity")
Traceback (most recent call last):
  File "<console>", line 1, in <module>
AttributeError: 'ShotgunWrapper' object has no attribute 'nav_expand'
```

Added to hot topics for discussion.

josh-t: Edited original response as it had a bad test case example.

If I call `nav_expand` with a newer version of python-api on old site, I get:

```python
>>> shotgun.server_caps.version
(6, 3, 9)
>>> shotgun.nav_expand("/", "Version.entity")
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/shotgun/configs/sgutils_test/install/core/python/tank_vendor/shotgun_api3/shotgun.py", line 2068, in nav_expand
    'entity_fields': entity_fields
  File "/shotgun/configs/sgutils_test/install/core/python/tank/authentication/shotgun_wrapper.py", line 59, in _call_rpc
    return super(ShotgunWrapper, self)._call_rpc(*args, **kwargs)
  File "/shotgun/configs/sgutils_test/install/core/python/tank_vendor/shotgun_api3/shotgun.py", line 2149, in _call_rpc
    self._response_errors(response)
  File "/shotgun/configs/sgutils_test/install/core/python/tank_vendor/shotgun_api3/shotgun.py", line 2425, in _response_errors
    raise Fault(sg_response.get("message", "Unknown Error"))
Fault: Uncaught exception! undefined method `nav_expand' for #<API3Handler:0x007f166622aca8>
>>> 
```

I agree, it feels like a proper interface or expected exception for checking supported features seems right. Like I said before, I've added to hot topics for discussion.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: Agree.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: My thought was that as long as `WARNING` showed up, it would be sufficient and most views would have scrolling/cell expansion to show the rest. I'm definitely open to change it though.

info.yml
josh-t: This removes the `7.0.2` requirement unless I'm misunderstanding your comment.

python/shotgun_model/shotgun_hierarchy_model.py
manneohrstrom: as long as it's readable it works for me! :)

info.yml
manneohrstrom: sorry i am an idiot. :(

PR #53, opened 2016-09-23
#37113 Reverts usage of some of the newer core features.

josh-t:
In order to fail gracefully on older versions of core, needed to revert usage of some of the features added in recent cores. Specifically, use of utilities like the `filesystem` and `logging` modules. 

Also made tweaks to provide better immediate feedback as to why hierarchy isn't supported.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -30,7 +30,6 @@ Contents:
    :maxdepth: 2
 
    shotgun_model
-   shotgun_hierarchy_model
    shotgun_data
    task_manager
    settings
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -5,9 +5,8 @@ The Shotgun Utils Framework contains a collection of helpers and tools to make i
 painless to build consistent looking applications. It contains several modules which can each
 be used independently:
 
-- Shotgun Data and Shotgun Hierarchy models deriving from
-  :class:`~PySide.QtGui.QStandardItemModel` which makes it easy to quickly build
-  responsive, rich applications with data from your Shotgun site.
+- Shotgun Data model deriving from :class:`~PySide.QtGui.QStandardItemModel` which makes 
+  it easy to quickly build responsive, rich applications with data from your Shotgun site.
 
 - A settings system handles per-user preferences and makes it easy to work with the
   :class:`~PySide.QtCore.QSettings` class from within your app code.
--- a/python/shotgun_model/shotgun_hierarchy_model.py
+++ b/python/shotgun_model/shotgun_hierarchy_model.py
@@ -83,10 +83,8 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
         super(ShotgunHierarchyModel, self).__init__(parent, bg_task_manager)
 
         # check for hierarchy support
-        current_engine = sgtk.platform.current_engine()
-        sg_connection = current_engine.shotgun
         (self._hierarchy_is_supported, self._hierarchy_not_supported_reason) = \
-            hierarchy_is_supported(sg_connection)
+            self.__hierarchy_is_supported()
 
         if not self._hierarchy_is_supported:
             self._log_warning(self._hierarchy_not_supported_reason)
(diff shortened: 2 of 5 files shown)


python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -961,16 +938,22 @@ def hierarchy_is_supported(sg_connection):
...
-        server_caps.version and
-        server_caps.version >= (7, 0, 2) and
-        hasattr(sg_connection, "nav_expand")
-    )
+    if not (hasattr(sg_connection, "server_caps") and
+            server_caps.version and
+            server_caps.version >= (7, 0, 2)):
+        return (False, "The version of SG being used does not support hierarchy queries. v7.0.2 is required.")
+    elif not hasattr(sg_connection, "nav_expand"):
+        return (False, "The verison of the python-api being used does not support hierarchy queries.")
```
jfboismenu: version

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -961,16 +938,22 @@ def hierarchy_is_supported(sg_connection):
...
-    return (
-        hasattr(sg_connection, "server_caps") and
-        server_caps.version and
-        server_caps.version >= (7, 0, 2) and
-        hasattr(sg_connection, "nav_expand")
-    )
+    if not (hasattr(sg_connection, "server_caps") and
+            server_caps.version and
+            server_caps.version >= (7, 0, 2)):
+        return (False, "The version of SG being used does not support hierarchy queries. v7.0.2 is required.")
```
jfboismenu: Maybe it's because I'm not a native english speaker, but I think "does not support querying for the hierarchy"

python/shotgun_model/shotgun_hierarchy_model.py
jfboismenu: Same comment about hierarchy queries.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -961,16 +938,22 @@ def hierarchy_is_supported(sg_connection):
 
     :param sg_connection: A shotgun connection.
 
-    :returns: ``True`` if hierarchy supported, ``False`` otherwise.
+    :returns: A tuple of 2 items where the first item is a boolean indicating
```
jfboismenu: I'm starting to use the `:rtype type:` and `:param type name: documentation`way of documenting stuff, I feel like the wording is better. I'm not sure this would help that much in this case, but I thought I'd mention it in case you wanted to experiment with it.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -684,20 +681,36 @@ def _load_cached_data(self):
...
         except Exception, e:
-            logger.debug(
+            self._log_debug(
                 "Couldn't load cache data from disk.\n"
                 " Will proceed with full SG load.\n"
                 "Error reported: %s" % (e,)
             )
             return False
 
+    def _log_debug(self, msg):
```
jfboismenu: Feels like a recipe that should be part of the framework, something like a class you could derive from that has all the _log_* methods we need and uses .log when available. Let's discuss this in dailies.

docs/index.rst
```diff
@@ -30,7 +29,6 @@ Contents:
    :maxdepth: 2
 
    shotgun_model
-   shotgun_hierarchy_model
```
jfboismenu: Why are you removing this from the doc? Not officially supported yet?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -85,7 +82,14 @@ def __init__(self, parent, schema_generation=0, bg_task_manager=None):
 
         super(ShotgunHierarchyModel, self).__init__(parent, bg_task_manager)
 
-        self._hierarchy_is_supported = self._check_hierarchy_is_supported()
+        # check for hierarchy support
+        current_engine = sgtk.platform.current_engine()
+        sg_connection = current_engine.shotgun
+        (self._hierarchy_is_supported, self._hierarchy_not_supported_reason) = \
+            hierarchy_is_supported(sg_connection)
```
jfboismenu: Wondering if hierarchy_is_supported can be a private class method instead of module public method. Is this used anywhere else?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -407,7 +381,10 @@ def _load_data(
         if not self._hierarchy_is_supported:
             self.clear()
             root = self.invisibleRootItem()
-            item = QtGui.QStandardItem("WARNING: Hierarchy not supported. See logged warnings for details.")
+            item = QtGui.QStandardItem("WARNING: Hierarchy not supported")
+            item.setEditable(False)
+            root.appendRow([item])
+            item = QtGui.QStandardItem("- %s" % (self._hierarchy_not_supported_reason,))
```
jfboismenu: Nice!

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: The demo app was using it, but I think you're right. I've updated based on your suggestion.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: I've been trying to use `:param type name:` as well when possible. I haven't been using `:rtype:` as much, but I like it. Updated here.

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: Updated to: `does not support querying for the project hierarchy.`

python/shotgun_model/shotgun_hierarchy_model.py
josh-t: good eye. fixed.

josh-t: fixed with same as above.

docs/index.rst
josh-t: Correct. The python-api method is also not official yet, so holding off for now.

PR #55, opened 2016-11-04
38712 clear globals cache

akhronek:
Added a class method to CachedShotgunSchema that removes schema and status cache files from disk for the current/specified project, if the files exist.

### DIFF
--- a/python/shotgun_globals/cached_schema.py
+++ b/python/shotgun_globals/cached_schema.py
@@ -1,11 +1,11 @@
 # Copyright (c) 2015 Shotgun Software Inc.
-# 
+#
 # CONFIDENTIAL AND PROPRIETARY
-# 
-# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+#
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit
 # Source Code License included in this distribution package. See LICENSE.
-# By accessing, using, copying or modifying this work you indicate your 
-# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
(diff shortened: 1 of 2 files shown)


python/shotgun_globals/cached_schema.py
```diff
@@ -881,3 +881,27 @@ def get_ordered_status_list(cls, display_names=False, project_id=None):
...
+                           None, the current context's project will
+                           be used.
+        """
+        self = cls.__get_instance()
+        project_id = project_id or self._get_current_project_id()
+
+        schema_cache = self._get_schema_cache_path(project_id)
+        if os.path.isfile(schema_cache):
+            self._bundle.log_debug("Removing schema cache file : %s" % schema_cache)
+            os.remove(schema_cache)
```
thebeeland: Let's log an error for these if they fail:

```
try:
    os.remove(schema_cache)
except Exception:
    self._bundle.log_error( ... )
    raise
```

Unfortunately, it looks like it will need to be `Exception` that we catch instead of `OSError`, as we get a `WindowsError` on Windows, but that exception doesn't exist for non-Windows OSs. Meh.

python/shotgun_globals/cached_schema.py
```diff
@@ -881,3 +881,27 @@ def get_ordered_status_list(cls, display_names=False, project_id=None):
...
+
+        status_cache = self._get_status_cache_path(project_id)
+        if os.path.isfile(status_cache):
+            self._bundle.log_debug("Removing status cache file : %s" % status_cache)
+            os.remove(status_cache)
```
thebeeland: Same comment as above!

PR #56, opened 2016-11-16
#39094 Support bubble field notation in sg globals methods.

josh-t:
Some methods that accept `sg_entity_type` and `field_name` arguments were already aware of "bubble" field notation (`sg_sequence.Sequence.code`). This change makes all these methods work when a field name is in "bubble" notation.

### DIFF
--- a/python/shotgun_globals/cached_schema.py
+++ b/python/shotgun_globals/cached_schema.py
@@ -551,8 +551,12 @@ class CachedShotgunSchema(QtCore.QObject):
 
         :returns: Field display name        
         """
+
         self = cls.__get_instance()
         project_id = project_id or self._get_current_project_id()
+
+        (sg_entity_type, field_name) = _account_for_bubble_fields(
+            sg_entity_type, field_name)
 
         self._check_schema_refresh(
             sg_entity_type,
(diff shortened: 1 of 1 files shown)


python/shotgun_globals/cached_schema.py
```diff
@@ -920,3 +967,29 @@ def clear_cached_data(cls, project_id=None):
                 )
                 raise
 
+
+def _account_for_bubble_fields(sg_entity_type, field_name):
+    """Detect bubble fields and return the proper entity type and field name.
+
+    :param str sg_entity_type: The intput entity type name. If the field name
```
manneohrstrom: Could this be added to core as well? That way, if we introduce the utility now, we can switch shotgun utils over at some point in the future. This feels like a really useful generic method that applies to any shotgun filter logic.

josh-t: Sure thing. Where do you think it should go? `tank/util/shotgun.py` or `tank/util/shotgun_entity.py` maybe?

manneohrstrom: Great! How about `shotgun.py`? then we can expose it as `sgtk.util.resolve_shotgun_bubble_fields()` or something like that - what do you think?

PR #57, opened 2016-11-16
#38475 Refactored shotgun model cache I/O backend

manneohrstrom:
This refactors the shotgun model cache backend and alters the behaviour of the shotgun model to enhance performance in scenarios where you are dealing with a lot of records.

### Design

image

- A new low level i/o layer is introduced. This is fully decoupled from the model level
- Each model data format (find vs nav_expand) is expressed as a subclass of the low level i/o `DataHandler`, implementing the specifics for how the shotgun data fetched should be turned into a cache dataset.
- A `DataHandler` instance currently corresponds to a single cache file on disk
- The Model level now knows very little about the data set. It merely reaches out to the `DataHandler` to request child data for a node
- Each model implements logic for how to create a model item given a `DataHandler` `DataItem`
- For domainspecific model logic (such as the sg model for example returning a list of all the shotgun ids currently held in the model), this is implemented in the subclassed `DataHandler`.
- Everything is deferred loaded into the model. The low level data handler keeps everything in memory and as you expand the model, more and more model items get constructed from the low level data.
- The performance of the new backend is orders of magnitude better than the previous implementation, especially on deep hierarchies.
- The nav/find implementations are much more aligned as a consequence of the separation of business logic into two layers.

### Backwards Compatibility

- The data handling now supports full and complete handling of update/delete/create scenarios for incoming data. Previously, if an updated value came in from shotgun, the treeview (in for example the loader) would fully reload and you would lose any selection state. This now works correctly, the view doesn't get reset and you merely see the value updated.
- Everything is loaded in on expand. This means that the model starts out with a subset of all items, and they are loaded in as you go. This may mean regressions in some search implementations that are searching for data in the model.
- I removed the nav API refresh as I think it needs a bit of discussion. @josh-t maybe we can discuss at some point what is best. I don't think we need to be compatible with the shotgun model in terms of interface - we should figure out what refresh means in context of the tree view.


python/shotgun_model/shotgun_query_model.py
```diff
@@ -620,262 +557,271 @@ def _set_tooltip(self, item, data):
...
+                    # assume that the top level is always loaded in tree
+                    # so that it's always safe to do data_item.parent.uid here
+                    parent_item = self._get_item_by_unique_id(data_item.parent.unique_id)
+                    # get model index
+                    parent_model_index = parent_item.index()
+                    # kick it
+                    self.fetchMore(parent_model_index)
+
+            # now try again
+            item = self._get_item_by_unique_id(uid)
```
josh-t: Is it true that this may return `None` still if the parent items are still being fetched?

python/shotgun_model/util.py
```diff
@@ -132,7 +132,7 @@ def sanitize_qt(val):
         return sanitize_qt(val)    
     
     elif isinstance(val, list):
-        return [ sanitize_qt(d) for d in val ]
+        return [sanitize_qt(d) for d in val ]
```
josh-t: remove space at end too. :)

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -1068,25 +1068,32 @@ def _on_task_completed(self, task_id, group, result):
             return
 
         action = result.get("action")
-        if action in ["find", "find_one", "create", "delete", "update"]:
-            self.work_completed.emit(str(task_id), action, {"sg":result["sg_result"]})
-        elif action == "nav_expand":
-            self.work_completed.emit(str(task_id), "nav_expand", {"nav":result["nav_result"]})
+        if action in ["find", "find_one", "create", "delete", "update", "nav_expand"]:
```
jfboismenu: `work_completed` used to emit `{"nav":result["nav_result"]}`but now emits `{"sg": result["sg_result"]}`, won't this break backwards compatibility?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -1068,25 +1068,32 @@ def _on_task_completed(self, task_id, group, result):
...
+                "method",
+                {"return_value": result["result"]}
+            )
         elif action == "check_thumbnail":
             path = result.get("thumb_path", "")
             if path:
                 # check found a thumbnail!
                 self.work_completed.emit(
                     str(task_id),
-                    "find",
```
jfboismenu: Same here. The payload is changing.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+    Borrowed from 0.18.
+    """
+    def wrapper(self, *args, **kwargs):
+        time_before = time.time()
+        try:
+            response = func(self, *args, **kwargs)
+        finally:
+            time_spent = time.time() - time_before
+            # log to special timing logger
+            self._bundle.log_debug("ShotgunDataHandler.%s took %fs" % (func.__name__, time_spent))
```
jfboismenu: This logger is builtin core 0.18. I guess @manneohrstrom  put it there because shotgunutils doesn't have that dependency yet?

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+
+    This class abstracts away the data management and allows
+    the model to access the data in a simple tree-like fashion.
+    Each node in the tree is also identified by a unique id
+    and can be accessed directly via this id in an O(1) lookup.
+
+    It also offers fast serialization and loading. Each
+    ShotgunDataHandler is connected to a single cache file on disk.
+
+    Each QT model typically sublcasses this in order to create
```
jfboismenu: Also, typo: sublcases -> subclasses

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+    # are made which renders the cache files non-backwards compatible.
+    FORMAT_VERSION = 27
+
+    # constants for updates
+    (UPDATED, ADDED, DELETED) = range(3)
+
+    # internal constants for serialization performance
+    (CACHE_BY_UID, CACHE_CHILDREN, IS_LEAF, UID, PARENT, FIELD, SG_DATA) = range(7)
+
+    def __init__(self, cache_path, parent):
```
jfboismenu: I think it's actually cleaner to specify the path to the file. It simplifies testing for one.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+        Create a string representation of this instance
+        :returns: A string representation of this instance
+        """
+        return "<%s@%s (%d items)>" % (
+            self.__class__.__name__,
+            self._cache_path,
+            len(self._cache[self.CACHE_BY_UID])
+        )
+
+    def _create_clear_cache(self):
```
jfboismenu: I would much prefer `_create_empty_cache`. Or at the very least `create_cleared_cache` which is (to my french eyes) looks better gramatically.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+        self._bundle = sgtk.platform.current_bundle()
+        # the path to the cache file
+        self._cache_path = cache_path
+        # data in cache
+        self._cache = None
+
+    def __repr__(self):
+        """
+        Create a string representation of this instance
+        :returns: A string representation of this instance
```
manneohrstrom: Hah! I think you actually wrote this code 😆  I agree i think it's OTT. I'll dial it down 👻

python/shotgun_model/data_handler.py
manneohrstrom: yep exactly. i'll clarify in a comment. Cannot wait for jan 2017.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+            response = func(self, *args, **kwargs)
+        finally:
+            time_spent = time.time() - time_before
+            # log to special timing logger
+            self._bundle.log_debug("ShotgunDataHandler.%s took %fs" % (func.__name__, time_spent))
+        return response
+    return wrapper
+
+
+class ShotgunDataHandler(QtCore.QObject):
```
manneohrstrom: hmm, that's a good question. I think at some point it emitted signals. But i don't think it does anymore. I'll see if i can just derive from object.

python/shotgun_model/data_handler.py
manneohrstrom: will fix!

python/shotgun_model/data_handler.py
manneohrstrom: i was torn myself. i think it was practical reasons in the end - i'll take another look and see if it makes sense to move it in.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+
+        :returns: new cache dictionary
+        """
+        return {
+            self.CACHE_CHILDREN: {},  # hierarchy
+            self.CACHE_BY_UID: {},    # uid-based lookup
+            self.UID: None            # the uid of the root is None
+        }
+
+    def _clear_cache(self):
```
manneohrstrom: In some cases you just need a stub cache object returned. This is what `_create_clear_cache` is for. i'll see if i can make the naming clearer.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+        :param unique_id:     Unique identifier, typically an int or a string
+        :param parent_object: Parent object that the requester wants to parent
+                              newly created nodes to. This object is passed into
+                              the node creation factory method as nodes are being
+                              created.
+        :param factory_fn:    Method to execute whenever a child node needs to
+                              be created. The factory_fn will be called with the
+                              following syntax: factory_fn(parent_object, data_item),
+                              where parent_object is the parent_object parameter and
+                              data_item is a :class:`ShotgunDataItem` representing the
```
manneohrstrom: `ShotgunData` sounds like a collection of things to me. Like what you get back from a `find()` call. Wrapper sounds to me like it's an adapter of sorts. 

I totally agree that it's confusing that there are model items and data items. 

How about `ShotgunDataNode` (because it's in a tree) or `ShotgunDatum`, `ShotgunRecord`. Actually, sadly, i think personally i think `item` sounds better than all the suggestions i could come up with. How strongly do you feel about this? In the code i tended to name the variables `data_item` and `model_item` which i thought ended up being reasonably clear.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+        elif isinstance(a, list):
+            if not isinstance(b, list):
+                return False
+            if len(a) != len(b):
+                return False
+            for idx in xrange(len(a)):
+                if not self._sg_compare_data(a[idx], b[idx]):
+                    return False
+
+        # handle thumbnail fields as a special case
```
manneohrstrom: do you know if we use it anywhere else right now?

python/shotgun_data/shotgun_data_retriever.py
manneohrstrom: ah! yes - good point, i wanted to bring this up and forgot (sorry code was written over an extended period of time). The nav data is brand new and not used anywhere. We do run the risk of breaking things but it's very low. This `nav` key makes things real hard - it's not how it was intended to be used - the `sg` key denotes that you are getting data back from Shotgun in some form or another - each action denotes what type of data is to be expected (find, create, delete, nav etc). I wish we could change the schema one too.

python/shotgun_data/shotgun_data_retriever.py
manneohrstrom: Yes - this is a cut n paste typo - @jfboismenu you ok with it changing to its "correct" value?

python/shotgun_model/data_handler.py
manneohrstrom: `_create_empty_cache` sounds great

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+    def is_cache_loaded(self):
+        """
+        Returns true if the cache has been loaded into memory, false if not.
+
+        :returns: boolean to indicate if cache is loaded
+        """
+        return self._cache is not None
+
+    @log_timing
+    def remove_cache(self):
```
jfboismenu: I agree with Josh. I find the state transitions confusing. The documentation is also confusing. By looking at the method's name and then the code I would have expected `is_cached_loaded` to return False after the call to `remove_cache`. However, this is not true, as `load_cache` is called right at the end.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+    def unload_cache(self):
+        """
+        Unloads any in-memory cache data.
+        """
+        if self._cache is None:
+            # nothing to do
+            return
+
+        self._log_debug("Unloading in-memory cache for %s" % self)
+        self._cache = None
```
jfboismenu: Managing a field inside the `_cache` dict or managing the `None`-ness of `_cache` feels the same to me. Am I missing something?

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+
+        if not self.is_cache_loaded():
+            return None
+
+        return ShotgunDataItem(self._cache[self.CACHE_BY_UID].get(unique_id))
+
+    @log_timing
+    def generate_child_nodes(self, unique_id, parent_object, factory_fn):
+        """
+        Generate nodes recursively from the data set
```
jfboismenu: missing period.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+
+        each node will be passed to the factory method for construction.
```
jfboismenu: Capital letter.

jfboismenu: Each

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+
+        unique id can be none, meaning generate the top level of the tree
+
+        :param unique_id:     Unique identifier, typically an int or a string
```
jfboismenu: Why is it strings or ints depending on the occasion? What does that imply? What are the consequences of that? Maybe this could be explained.

jfboismenu: After reading code, `None` is also a possibility.

python/shotgun_model/data_handler.py
josh-t: ah cool!

python/shotgun_model/data_handler.py
josh-t: Hilarious. I just spit some of my drink out. Thanks for dialing down my OTT. 😜

python/shotgun_model/data_handler.py
josh-t: Wrapping the cache in a little private object might make the code a little cleaner to read. Then this code wouldn't need to manage None vs. dict etc. The path logic, clear/remove/unload logic could live on that object. Maybe that's OTT though. Apparently I sway toward OTT sometimes without realizing it.

python/shotgun_model/data_handler.py
josh-t: One thing I thought of was just switching it around to `ShotgunItemData` or `ShotgunModelItemData`. That way the `Item` part is model item, and this is the data that is stored there (sort of)?

I'm not terribly passionate about it, but I did struggle to keep them separated in my head. Maybe this would help?

python/shotgun_model/data_handler.py
josh-t: haha... i was just looking at the previous conditionals which are totally generic and then *boom* we're in `amazonaws` land. probably fine to leave it for now.

tests/test_shotgun_model_data_handler.py
```diff
@@ -0,0 +1,144 @@
+# Copyright (c) 2013 Shotgun Software Inc.
```
jfboismenu: Did you forget to push your unit tests? :)

python/shotgun_data/shotgun_data_retriever.py
jfboismenu: If it's not used anywhere, by all means, change it.

python/shotgun_model/data_handler.py
manneohrstrom: done

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+            time_spent = time.time() - time_before
+            # log to special timing logger
+            self._bundle.log_debug("ShotgunDataHandler.%s took %fs" % (func.__name__, time_spent))
+        return response
+    return wrapper
+
+
+class ShotgunDataHandler(QtCore.QObject):
+    """
+    Abstract class that manages low level data storage for QT models.
```
manneohrstrom: done

python/shotgun_model/data_handler.py
manneohrstrom: i think the `load_cache` call was a mistake! I have removed it (and it's now unit tested).

python/shotgun_model/data_handler.py
manneohrstrom: renamed for clarity

python/shotgun_model/data_handler.py
manneohrstrom: hah! done!

python/shotgun_model/data_handler.py
manneohrstrom: unit tests did end up being easy this way. I kept it the way it is for the time being. we can always move it later.

python/shotgun_model/data_handler.py
manneohrstrom: done

python/shotgun_model/data_handler.py
manneohrstrom: i'll make a ticket (see comment below)

python/shotgun_data/shotgun_data_retriever.py
manneohrstrom: cool

python/shotgun_model/data_handler.py
manneohrstrom: i have clarified

python/shotgun_model/data_handler_nav.py
```diff
@@ -0,0 +1,230 @@
...
+            if self._SG_PATH_FIELD not in sg_item:
+                # note: leaf nodes of kind 'empty' don't have a path
+                unique_field_value = "/".join(parent_item[self.UID], sg_item["label"])
+
+            else:
+                unique_field_value = sg_item.get(self._SG_PATH_FIELD)
+
+
+            # this is an actual entity - insert into our new tree
+            item = {
```
manneohrstrom: yes, exactly!

tests/test_shotgun_model_data_handler.py
manneohrstrom: I had totally forgotten that i had started the tests a while back. Thanks for reminding. There is now a half-decent test rig in place for shotgun utils (yay!). It's light weight and can be a recipe for other frameworks to implement our standard test rig. It ended up being pretty neat actually :)

Tests added for the low level data handling.

python/shotgun_model/data_handler.py
manneohrstrom: i think i'm with JF on this one. `None` for me indicates that something is un-initialized or not present. The data handler has two states, loaded or unloaded, and in the unloaded state the cache is set to None. The methods load, unload and save (and remove) control the state of the data handler. I agree that the data structure is a little confusing, it's intentionally raw so that it is straight forward to pickle out to disk.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+        """
+        Saves the current cache to disk.
+        """
+        self._log_debug("Saving to disk: %s" % self)
+
+        # try to create the cache folder with as open permissions as possible
+        cache_dir = os.path.dirname(self._cache_path)
+
+        # make sure the cache directory exists
+        # todo: upgrade to 0.18 filesystem methods
```
manneohrstrom: yes! done!

python/shotgun_model/data_handler.py
manneohrstrom: I have renamed `ShotgunDataNode` -> `ShotgunItemData`

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+
+        for deleted_uid in current_uids.difference(new_uids):
+            item = self._cache[self.CACHE_BY_UID][deleted_uid]
+            diff_list.append({
+                "data": ShotgunDataItem(item),
+                "mode": self.DELETED
+            })
+            num_deletes += 1
+
+        # lastly swap the new for8 the old
```
manneohrstrom: done!

python/shotgun_model/data_handler_nav.py
```diff
@@ -0,0 +1,230 @@
...
+        num_deletes = 0
+        num_modifications = 0
+
+        # insert the new items in this dict
+        new_items = {}
+
+        # analyze the incoming shotgun data
+        for sg_item in sg_data["children"]:
+
+
```
manneohrstrom: done

python/shotgun_model/data_handler_nav.py
```diff
@@ -0,0 +1,230 @@
...
+            if self._SG_PATH_FIELD not in sg_item:
+                # note: leaf nodes of kind 'empty' don't have a path
+                unique_field_value = "/".join(parent_item[self.UID], sg_item["label"])
+
+            else:
+                unique_field_value = sg_item.get(self._SG_PATH_FIELD)
+
+
```
manneohrstrom: done

python/shotgun_model/data_item.py
```diff
@@ -0,0 +1,79 @@
...
+# By accessing, using, copying or modifying this work you indicate your
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+from .data_handler import ShotgunDataHandler
+
+
+class ShotgunDataItem(object):
+    """
+    Object wrapper around an entry in the :class:`DataHandler`.
```
manneohrstrom: Great suggestion! i'll add a big @todo for us to hit in the future :)

tests/fixtures/config/bundles/test_app/app.py
```diff
@@ -0,0 +1,22 @@
...
+# CONFIDENTIAL AND PROPRIETARY
+# 
+# This work is provided "AS IS" and subject to the Shotgun Pipeline Toolkit 
+# Source Code License included in this distribution package. See LICENSE.
+# By accessing, using, copying or modifying this work you indicate your 
+# agreement to the Shotgun Pipeline Toolkit Source Code License. All rights 
+# not expressly granted therein are reserved by Shotgun Software Inc.
+
+"""
+A dummy app
```
jfboismenu: I don't think it's very nice to call an app a dummy. :p

tests/shotgun_model/test_data_handler.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # remove from disk
+        dh.remove_cache()
+
+        # no cache file on disk
+        self.assertFalse(dh.is_cache_available())
+
+        # but it is loaded
+        self.assertFalse(dh.is_cache_loaded())
+
```
jfboismenu: Nit pick: extra lines.

tests/shotgun_model/test_data_handler_find.py
```diff
@@ -0,0 +1,247 @@
...
+            parent=None
+        )
+
+        dh.load_cache()
+        self.assertEquals(len(dh.get_entity_ids()), 0)
+
+        # first let the data handler perform its request
+        # and make sure it registers it in an expected way
+        mock_data_retriever = Mock()
+        mock_data_retriever.execute_find = Mock(return_value=1234)
```
jfboismenu: `Mock` is definitely the coolest Python tool I've discovered since I've started working on Toolkit. I may however be guilty of over-engineering some tests because of it. :)

tests/shotgun_model/test_data_handler_find.py
```diff
@@ -0,0 +1,247 @@
...
+            hierarchy=["sg_asset_type", "code"],
+            fields=["code"],
+            download_thumbs=True,
+            limit=None,
+            additional_filter_presets=None,
+            cache_path=test_path,
+            parent=None
+        )
+
+        dh.load_cache()
```
jfboismenu: Nitpick: All your tests start with

```python
dh = ShotgunFinDataHandler(...)
dg.load_cache()
```

To me that's a sign that maybe the cache should be lazy loaded by update operations. Otherwise the API is error prone. It's an internal API with very limited scope, so it's not that big of a deal. Curious what you think.

python/shotgun_model/data_handler_nav.py
```diff
@@ -0,0 +1,230 @@
...
+                self.FIELD: None,
+                self.IS_LEAF: not sg_item["has_children"],
+                self.UID: unique_field_value,
+                self.PARENT: parent_item,
+                self.CACHE_CHILDREN: {},
+            }
+
+            # if we have children in the existing cache, include those
+            if unique_field_value in self._cache[self.CACHE_BY_UID]:
+                item[self.CACHE_CHILDREN] = self._cache[self.CACHE_BY_UID][unique_field_value][self.CACHE_CHILDREN]
```
manneohrstrom: the cache is now in a separate class

python/shotgun_model/shotgun_query_model.py
```diff
@@ -620,262 +557,271 @@ def _set_tooltip(self, item, data):
...
+        # this node is loaded in the query cached by the data handler
+        # but may not exist in the model yet - because of deferred loading.
 
-    def _sg_clean_data(self, sg_data):
-        """
-        Recursively clean the supplied SG data for use by clients.
+        # first see if we have it in the model already
+        item = self._get_item_by_unique_id(uid)
+
+        if not item:
```
manneohrstrom: i don't like returning early, it can get confusing too. indent isn't too bad, less than one page long, so fairly readable. i'll leave it as is.

python/shotgun_model/shotgun_query_model.py
manneohrstrom: ah yes, good point. This is actually a separate bug - if you are running the loader for example, the very first time, when the cache isn't primed, the home button will not take you to the right location by default, because the data isn't loaded yet. this is already an issue and i think we have a separate ticket for it!

python/shotgun_model/data_handler.py
manneohrstrom: now gone completely.

python/shotgun_model/data_handler.py
```diff
@@ -0,0 +1,473 @@
...
+        Both inputs are assumed to contain utf-8 encoded data.
+
+        :returns: True if a is same as b, false otherwise
+        """
+        if isinstance(a, dict):
+            if not isinstance(b, dict):
+                return False
+            if len(a) != len(b):
+                return False
+            for a_key in a.keys():
```
manneohrstrom: done

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+        :rtype: ``list``
+        """
+        # loop over all cache items - the find data handler organizes
+        # its unique ids so that all leaf nodes (e.g. representing an entity
+        # are ints and all other items are strings
+
+        # memoized for performance
+        if self.__entity_ids is None:
+            entity_ids = []
+            for uid in self._cache[self.CACHE_BY_UID].keys():
```
manneohrstrom: changed

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+
+    def get_uid_from_entity_id(self, entity_id):
+        """
+        Returns the unique id for a given entity or None if not found
+
+        :param entity_id: Shotgun entity id to resolve
+        :returns: unique id as string or int, to be used with
+                  :meth:`get_data_item_from_uid`.
+        """
+        for uid in self._cache[self.CACHE_BY_UID].keys():
```
manneohrstrom: changed

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+        Generate a data request for a data retriever.
+
+        Once the data has arrived, the caller is expected to
+        call meth:`update_data` and pass in the received
+        data payload for processing.
+
+        :param data_retriever: :class:`~tk-framework-shotgunutils:shotgun_data.ShotgunDataRetriever` instance.
+        :returns: Request id or None if no work is needed
+        """
+        # only request data from shotgun is filters are defined.
```
manneohrstrom: you are confusing `filters=[]` with `filters=None` - setting filters to `None` is an indication that we don't want any data to be fetched but the model to be populated with external data only!

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+        if self.__filters is None:
+            request_id = None
```
manneohrstrom: fixed

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+
+        else:
+            # get data from shotgun - list/set cast to ensure unique fields
+            fields = self.__hierarchy + self.__fields
```
manneohrstrom: gosh, this code is really old. I am not sure. I think its just trying to generate a list of unique items, that's all.

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+        :returns: list of updates. see above
+        :raises: :class:`ShotgunModelDataError` if no cache is loaded into memory
+        """
+        self._log_debug("Updating %s with %s shotgun records." % (self, len(sg_data)))
+        self._log_debug("Hierarchy: %s" % self.__hierarchy)
+
+        if self._cache is None:
+            raise ShotgunModelDataError("No data currently loaded in memory!")
+
+        if len(self._cache[self.CACHE_CHILDREN]) == 0:
```
manneohrstrom: replaced by low level cache wrapper class.

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+                    # get the parent uid or None if we are at the root level
+                    parent_uid = sub_tree.get(self.UID)
+                    # generate path for this item
+                    unique_field_value = self.__generate_unique_key(parent_uid, field_name, sg_item)
+                else:
+                    # on the leaf level, use the entity id as the unique key
+                    unique_field_value = sg_item["id"]
+
+                # two distinct cases for leaves and non-leaves
+                if on_leaf_level:
```
manneohrstrom: this code is already fairly complex. i am worried we will make it harder to read if we do what you suggest.

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+                                })
+                                num_modifications += 1
+
+                    # recurse down to the next level
+                    sub_tree = sub_tree[self.CACHE_CHILDREN][unique_field_value]
+
+        # now figure out if anything has been removed
+        self._log_debug("Diffing new tree against old tree...")
+
+        current_uids = set(self._cache[self.CACHE_BY_UID].keys())
```
manneohrstrom: changed

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+        new_uids = set(new_cache[self.CACHE_BY_UID].keys())
+
+        for deleted_uid in current_uids.difference(new_uids):
```
manneohrstrom: not sure - the data sets are usually very similar in size, i don't think it makes a practical difference. Memory wise i think it's the same.

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+                "mode": self.DELETED
+            })
+            num_deletes += 1
+
+        # lastly swap the new for8 the old
+        self._clear_cache()
+
+        # at this point, kick the gc to make sure the memory is freed up
+        # despite its cycles.
+        gc.collect()
```
manneohrstrom: it seemed to help free up memory. I ran some quick and fairly unscientific tests :)

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+        # and thus affect memory usage and i/o peformance. We assume
+        # that a vast majority of nodes are leaves and store these as
+        # ints for compactness. non-leaf nodes have a "path-like" string
+        # to uniquely describe their location in the tree.
+        #
+        # we assume that on each level, values are unique.
+
+        value = sg_data.get(field)
+
+        if isinstance(value, dict) and "id" in value and "type" in value:
```
manneohrstrom: i think if you have a polymorphic join, there is a problem here. thanks for the heads up!

so for example if you show tasks grouped by link, you could have tasks for assets and shots in the same list, generating the same ids!

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+            # for single entity links, return the entity id
+            unique_key = value["id"]
+
+        elif isinstance(value, list):
+            # this is a list of some sort. Loop over all elements and extract a comma separated list.
+            formatted_values = []
+            if len(value) == 0:
+                # no items in list
+                formatted_values.append("_")
+            for v in value:
```
manneohrstrom: done

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+        if isinstance(value, dict) and "id" in value and "type" in value:
+            # for single entity links, return the entity id
+            unique_key = value["id"]
+
+        elif isinstance(value, list):
+            # this is a list of some sort. Loop over all elements and extract a comma separated list.
+            formatted_values = []
+            if len(value) == 0:
+                # no items in list
+                formatted_values.append("_")
```
manneohrstrom: not sure. old code. i removed it.

python/shotgun_model/data_handler_find.py
```diff
@@ -0,0 +1,384 @@
...
+            for v in value:
+                if isinstance(v, dict) and "id" in v and "type" in v:
```
manneohrstrom: that seems overkill to me! i'm leaving it as it is.

python/shotgun_model/data_handler_nav.py
```diff
@@ -0,0 +1,230 @@
...
+            raise ShotgunModelDataError(
+                "Unexpected error occurred. Could not determine the path"
+                "from the queried hierarchy item."
+            )
+
+        if len(self._cache[self.CACHE_CHILDREN]) == 0:
+            self._log_debug("In-memory cache is empty.")
+
+        # ensure the data is clean
+        self._log_debug("sanitizing data...")
```
manneohrstrom: i thought about it - it's actually not that much commonality - mostly init stuff, so i left it separate. but i wasn't sure.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -239,6 +205,12 @@ def clear(self):
             # a crash in some versions of shiboken
             # (see https://bugreports.qt-project.org/browse/PYSIDE-158 )
             self.__do_depth_first_tree_deletion(self.invisibleRootItem())
+
+            # unload the data backend
+            if self._data_handler:
+                self._data_handler.unload_cache()
```
manneohrstrom: this is cleared up by the new low level cache class!

python/shotgun_model/shotgun_query_model.py
```diff
@@ -318,122 +287,117 @@ def reset(self):
             "email support@shotgunsoftware.com."
         )
 
-    ############################################################################
-    # protected properties
-
-    def _get_cache_path(self):
+    def hasChildren(self, index):
```
manneohrstrom: not sure. a lot of the code you are commenting on is stuff that i have moved around, i was wondering the same thing as i was shifting it :) I think @josh-t wrote it that way to be resilient, which seems like a nice approach to me.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -318,122 +287,117 @@ def reset(self):
...
+        :type index: :class:`~PySide.QtCore.QModelIndex`
         """
-        return self.__full_cache_path
+        if not index.isValid():
+            return super(ShotgunQueryModel, self).hasChildren(index)
+
+        item = self.itemFromIndex(index)
+
+        if not isinstance(item, ShotgunStandardItem):
+            return super(ShotgunQueryModel, self).hasChildren(index)
```
manneohrstrom: i think you could inject other types of items into the model possibly. i'll add a comment. again, this is old code.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -889,131 +835,168 @@ def __do_depth_first_tree_deletion(self, node):
         for index in range(node.rowCount())[::-1]:
```
manneohrstrom: adjusted.

tests/shotgun_model/test_data_handler.py
```diff
@@ -0,0 +1,85 @@
...
+    def test_basic_io(self):
+        """
+        Test loading and saving
+        """
+        test_path = os.path.join(self.tank_temp, "test_basic_io.pickle")
+
+        dh = self.shotgun_model.data_handler.ShotgunDataHandler(test_path, None)
+
+        # no cache file on disk
+        self.assertFalse(dh.is_cache_available())
```
manneohrstrom: thanks for the heads up! That's useful to know! I'll adjust

tests/shotgun_model/test_data_handler_find.py
manneohrstrom: i think the reasoning for this so that we could provide separate timing and logging for the cache load. But maybe it's overkill. I like the symmetry of a load/unload too. I might keep it for the time being and then we can see in the future if we should refactor.

python/shotgun_model/data_handler_cache.py
```diff
@@ -0,0 +1,198 @@
...
+
+    def get_child_uids(self, parent_uid):
+        """
+        Returns all the child uids for the given parent
+
+        :param parent_uid: Parent uid
+        :returns: list of child uids
+        """
+        if parent_uid is None:
+            return self._cache[self.CACHE_CHILDREN].keys()
```
jfboismenu: It would be nice if get_child_uids returned an iterable object (.iterkeys()), just like your other get_multiple_somethings bellow. It would make the interface more consistent. The caller creates a set from that data anyway.

python/shotgun_model/data_handler_cache.py
```diff
@@ -0,0 +1,198 @@
...
+        The number of items in the cache
+        """
+        return len(self._cache[self.CACHE_BY_UID])
+
+    @property
+    def uids(self):
+        """
+        All the uids in this cache as a set
+        """
+        return set(self._cache[self.CACHE_BY_UID].iterkeys())
```
jfboismenu: Might as well return the iterator as itself and have the caller create the set. Then your interface is consistent collection management-wise.

Btw, I'm being nitpicky here only because you decided to make this interface awesome by yielding and returning iterators. :)

python/shotgun_model/data_handler_cache.py
```diff
@@ -0,0 +1,198 @@
+# Copyright (c) 2016 Shotgun Software Inc.
```
jfboismenu: Thanks for abstracting out this piece of code. It really makes the model simpler to read! :)

python/shotgun_model/data_handler_nav.py
```diff
@@ -30,7 +31,7 @@ class ShotgunNavDataHandler(ShotgunDataHandler):
     _SG_PATH_FIELD = "path"
     _SG_PARENT_PATH_FIELD = "parent_path"
 
-    def __init__(self, root_path, seed_entity_field, entity_fields, cache_path, parent):
+    def __init__(self, root_path, seed_entity_field, entity_fields, cache_path):
```
jfboismenu: Why was parent removed? You'll run the risk of having Qt delete the model at a later date and cause a crash or  memory leak in certain DCCs, as evidenced by the work I had to put it for releasing wf2.

python/shotgun_model/util.py
```diff
@@ -31,6 +33,7 @@ def get_sg_data(item):
     from .shotgun_model import ShotgunModel
     return get_sanitized_data(item, ShotgunModel.SG_DATA_ROLE)
 
+
```
jfboismenu: PEP8ing I see :)

python/shotgun_model/data_handler_cache.py
manneohrstrom: yeah, totally agree! Just so hard to know when to stop refactoring... :) This makes it a lot clearer which is great

python/shotgun_model/util.py
manneohrstrom: pycharm is doing all the work for me... 😆

python/shotgun_model/data_handler_cache.py
manneohrstrom: yep, hah, already did exactly this change when i resolved your previous comment.

python/shotgun_model/data_handler_nav.py
manneohrstrom: ah! as per Josh's suggestion, the data handler no longer inherits from qobject. it's just a normal py-class.

manneohrstrom:
YES! FINALLY!

jfboismenu:
Ship it!

PR #58, opened 2017-01-05
#37239: Use the new tree widget for Loader hierarchy navigation.

fpainchaud:
Add convenience method `SimpleShotgunHierarchyModel.reload_data()` that reloads Shotgun data into the model using the latest parameter values passed to `load_data()`.

This new method is needed for my updates to the Loader.

### DIFF
--- a/python/shotgun_model/simple_shotgun_hierarchy_model.py
+++ b/python/shotgun_model/simple_shotgun_hierarchy_model.py
@@ -63,3 +63,15 @@ class SimpleShotgunHierarchyModel(ShotgunHierarchyModel):
             path=path,
             entity_fields=entity_fields
         )
+
+    def reload_data(self):
+        """
+        Convenience method that reloads Shotgun data into the model
+        using the latest parameter values passed to :meth:`load_data()`.
+        """
+
+        super(SimpleShotgunHierarchyModel, self)._load_data(
+            self._seed_entity_field,
+            path=self._path,
+            entity_fields=self._entity_fields
+        )


josh-t:
I'm curious what the use case is for reloading the model with the same seed. Code itself looks fine, just want to make sure we're on the same page.

fpainchaud:
The current tab views implement a right click menu action "Refresh" that calls `model.async_refresh()`.

As Manne put it in a chat we had earlier this week, _the concept of a “refresh” for the hierarchy model is so ambigous_. For the hierarchy, we agreed to _clear the model/view and then reload it_. So, the new hierarchy tab views implement a right click menu action "Reload" that calls `model.reload_data()`.

manneohrstrom:
The concept of a "reload" in the cached hierarchy data set is complicated. There are two ways we can do it:

- We look at what has been loaded into the model and figure out all the sg queries needed to refresh the currently loaded data chunks. If you have been browsing around for a while in for example the loader, this could result in hundreds of queries sent to the sg server. This was the original implementation but i backed away from that as part of the refactor as I see it as a high risk thing. Maybe i am overly cautious, but i don't want a hierarchy model to accidentally start flooding the servers with requests because someone has does a big refresh. It's also confusing when compared with the shotgun model in which a refresh results in a single query. It's not obvious from the outside that a simple "refresh" would result in all those requests.

- The other approach would be to basically collapse the tree and force the user to traverse it again. as part of the manual traverse, nodes are then expanded again, data is loaded in and requested from shotgun.

Looking at the GMA, i don't see a way to refresh that tree view right now. The way to do it is to basically refresh the browser, which will collapse the tree, preserve the selection that you previously had and expand down to that. See picture:

image

So from a user's point of view, the client side data model in the GMA does not have a concept of reload. The code that is executed on refresh is simply to reinstate the fully collapsed model and then expand it along the path of the current selection until revealed.

To keep things simple, i don't think our cached shotgun hierarchy model should have a concept of refresh either. What does it even mean? I think the term hints a preservation of the current state but where we ensure that the data is up to date - which is what the first solution above is doing, but it's expensive and complex.

I think we need a slightly different term for this. Basically what the loader in this case needs to do i guess is:

1. tell the model to clear itself and unload anything it has got in memory. Any references kept to any model items are given up since these become orphaned and invalid.
2. as a consequence, all connected views stop showing any items
3. root level items are loaded into the model again
4. user starts traversing the tree, just like what happened at app startup
5. optionally we may want to preserve selection state by recursively load up the tree like in shotgun

Does that make sense?

We want a name that indicates to the user that we will be clearing the data and start over again. I was looking at if `clear()` would do this, and it looks like the QT concept of clear basically empties the entire model. I guess we want to clear and then have the model show the root level of items available. 

Maybe `clear_and_load()` ? Not totally in love with this, but cannot think of anything better.

What do you guys think?

josh-t:
Could we do something like a `reload_item()` for the hierarchy model (essentially clear only the chlidren for a given item and call `_request_data(path)`). Displayed in the UI as a **Refresh Children** or **Refresh Subtree** context menu option. I think when people do a reload they have some expectation of what level they're looking for something to be updated. If not, they could always run this when the top-level item is selected.

So basically allow a reload the same way @manneohrstrom suggests, but give the user control over what level the reload should happen. We could just leave the results collapsed or do the selection preservation with auto-expansion given the items still match. 

Does that make any sense?

manneohrstrom:
Just to follow up and include some of the off line conversation:

hey guys! I just spoke to Josh about this and we reckon the simplest way forward may just be to skip the refresh functionality in the loader for hierarchy views for the time being.

It looks like it is currently needed for a right click menu in the left-hand-side tree view - this one we can get rid of for the hierarchy widget.

The other one seems to happen when you click a tab - the loader tries to refresh the data at that point. same here, i think for the hierarchy view we can skip that.

Are there any other places @fpainchaud?

Potentially we could implement the right click refresh() action to basically clear the model and set it again (without changes to the shotgunutils fw) - i think for now this can all be handled on the loader side. Hope this helps - ping any of us if you have further questions!

fpainchaud:
If you look at my Loader CR, you will see that already (1) the hierarchy model is not refreshed/reloaded when switching tabs, and (2) the action menu implements a "Reload" for hierarchy tabs. These are the only 2 "refresh" cases.

robblau:
Hey all, just caught up on the conversation.

Waiting sounds good to me.  In terms of what I'd love to see when refresh is done is that post-refresh the tree retains its state as much as possible.  So if something is expanded before it is after (if it still exists).  New nodes are collapsed.  If the selected item is still there, then it stays selected.  Otherwise the closest descendant that still exists is selected.  Scroll should keep selection as in place as possible.

If we're punting this, do we need a ticket to track reload?  I can either create it or add the above to it.

fpainchaud:
@robblau A new ticket should be created for the behavior you described.

For the current ticket, I moved my `reload_data()` convenience method into a new class derived from `SimpleShotgunHierarchyModel` and living in the loader.

I will close this PR and delete this branch.

PR #59, opened 2017-01-11
For #40080, support non- .jpeg thumbnail filetypes

akhronek:
Removed hard-coded '.jpeg' extension for downloaded thumbnails. Let sgtk.util.downolad_url() determine the file extension from the (possibly redirected) url.

### DIFF
--- a/python/shotgun_data/shotgun_data_retriever.py
+++ b/python/shotgun_data/shotgun_data_retriever.py
@@ -180,7 +180,8 @@ class ShotgunDataRetriever(QtCore.QObject):
             # proxy and connection settings as set in the SG API are used
             try:
                 # Ask sgtk.util.download_url() to append the file type extension
-                # to the input filebase path to get the full path to the cache file.
+                # to the input path_to_cached_thumb to get the full path to the
+                # cache file.
                 full_path = sgtk.util.download_url(
                     bundle.shotgun, url, path_to_cached_thumb, True
                 )
@@ -723,7 +724,7 @@ class ShotgunDataRetriever(QtCore.QObject):
             # proxy and connection settings as set in the SG API are used
             try:
                 # Ask sgtk.util.download_url() to append the file type extension
-                # to the input filebase path to get the full path to the cache file.
+                # to the input file_path to get the full path to the cache file.
                 download_path = sgtk.util.download_url(
                     self._bundle.shotgun, url, file_path, True
                 )
@@ -762,7 +763,7 @@ class ShotgunDataRetriever(QtCore.QObject):
                     url = sg_data[field]
                     try:
                         # Ask sgtk.util.download_url() to append the file type extension
-                        # to the input filebase path to get the full path to the cache file.
+                        # to the input file_path to get the full path to the cache file.
                         download_path = sgtk.util.download_url(
                             self._bundle.shotgun, url, file_path, True
                         )
--- a/python/shotgun_data/shotgun_data_retriever.py
+++ b/python/shotgun_data/shotgun_data_retriever.py
@@ -1149,9 +1149,7 @@ class ShotgunDataRetriever(QtCore.QObject):
             file_path=file_path,
         )
 
-    def _task_download_thumbnail(
-            self, thumb_path, url, entity_type, entity_id, field, load_image, **kwargs
-        ):
+    def _task_download_thumbnail(self, thumb_path, url, entity_type, entity_id, field, load_image, **kwargs):
         """
         Download the thumbnail for the specified entity type, id and field.  This downloads the
         thumbnail into the thumbnail cache directory and returns the cached path.
--- a/python/shotgun_data/shotgun_data_retriever.py
+++ b/python/shotgun_data/shotgun_data_retriever.py
@@ -173,9 +173,6 @@ class ShotgunDataRetriever(QtCore.QObject):
 
             # download using standard core method. This will ensure that
             # proxy and connection settings as set in the SG API are used
-            # This method will also determine the correct file extension to
-            # download to, as that information may not be readily available
-            # from the url.
             sgtk.util.download_url(bundle.shotgun, url, path_to_cached_thumb)
 
             # modify the permissions of the file so it's writeable by others
@@ -220,8 +217,9 @@ class ShotgunDataRetriever(QtCore.QObject):
             # download using standard core method. This will ensure that
             # proxy and connection settings as set in the SG API are used
             # Also force this method to determine the correct file extension
-            # to download to, since that information cannot be parsed from
-            # the thumbnail source url.
+            # to download to, since that information cannot be determined from
+            # the thumbnail source url, typically in the form of:
+            # https://my-site.shotgunstudio.com/thumbnail/full/Asset/2929
             try:
                 path_to_cached_thumb = sgtk.util.download_url(
                     bundle.shotgun, thumb_source_url, path_to_cached_thumb, True
@@ -650,13 +648,6 @@ class ShotgunDataRetriever(QtCore.QObject):
                   work_completed and work_failure signals, making it
                   possible to match them up.
         """
-        if not self._task_manager:
-            self._bundle.log_warning(
-                "No task manager has been associated with this data retriever. "
-                "Unable to request thumbnail."
-            )
-            return
-
         # construct the url that refers to the thumbnail's source image
         thumb_source_url = urlparse.urlunparse((
             bundle.shotgun.config.scheme, bundle.shotgun.config.server,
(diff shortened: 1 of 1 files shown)


python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -689,14 +829,23 @@ def _get_thumbnail_path(url, bundle, directory_only=False):
         # If we were only asked to give back a directory path then we can
         # skip building and appending a file name.
         if not directory_only:
-            path_chunks.append("%s.jpeg" % hash_str[4:])
+            path_chunks.append(
+                "%s%s" % (hash_str[4:], ".jpeg" if use_default_extension else "")
```
manneohrstrom: hmm, looking at both the new and the old code, what happens when `directory_only` is `True`? Looks like `hash_str[4:]` is dropped which cannot be correct?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -651,7 +790,7 @@ def _get_attachment_path(attachment_entity, bundle):
         return os.path.join(directory_path, file_name)
 
     @staticmethod
-    def _get_thumbnail_path(url, bundle, directory_only=False):
+    def _get_thumbnail_path(url, bundle, directory_only=False, use_default_extension=True):
```
manneohrstrom: This is confusing to me. What if we instead of adding `use_default_extension` simply changed the return value to be `(path, extension)` and in the case when the extension is known, it is returned? That seems more transparent to me.

Then we didn't have to pass this `use_default_extension` flag around everywhere.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -182,7 +184,63 @@ def download_thumbnail(url, bundle):
...
+    @staticmethod
+    def download_thumbnail_source(entity_type, entity_id, bundle):
+        """
+        Convenience and compatibility method for quick and easy synchronous thumbnail download.
+        This will retrieve the source file for a thumbnail given a shotgun entity type and id.
+        If the resolved thumbnail source file has already been cached, a path to it will be
+        returned instantly. Otherwise, it will be downloaded from Shotgun and placed in the
+        standard cache location on disk. The full path to cached thumbnail is returned. 
+
+        This is a helper method meant to make it easy to port over synchronous legacy
```
manneohrstrom: nitpick: I suggest we add something here and to the old `download_thumbnail` method to point out that with the `download_thumbnail` you always get a jpeg with a fixed resolution and with the `download_thumbnail_source` you get whatever image (res and format) that was uploaded to sg.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -596,29 +680,51 @@ def _download_url(self, file_path, url, entity_type, entity_id, field):
...
         """
         try:
+            download_path = sgtk.util.download_url(
+                self._bundle.shotgun, url, file_path, True
+            )
+            file_path = download_path
+        except TypeError:
+            # Probably an older version of core that doesn't have the
+            # sgtk.util.download_url() `use_url_exension` arg implemented.
+            # Try running without that arg value
```
manneohrstrom: pls clarify what 'that arg value' is more specifically

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -596,29 +680,51 @@ def _download_url(self, file_path, url, entity_type, entity_id, field):
...
                           where the fresh url is pulled from.
+        :returns: Full path the downloaded file. This value may be different
+                  than the input `file_path` if the resolved url's extension
+                  differed from what was specified.
         """
         try:
+            download_path = sgtk.util.download_url(
+                self._bundle.shotgun, url, file_path, True
+            )
+            file_path = download_path
```
manneohrstrom: nice!

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -596,29 +680,51 @@ def _download_url(self, file_path, url, entity_type, entity_id, field):
...
+        except TypeError:
+            # Probably an older version of core that doesn't have the
```
manneohrstrom: i am not 100% sure, but looks like the exception bubbling is not right - can you double check?

before, it was

```
try:
    sgtk.util.download_url(self._bundle.shotgun, url, file_path)
except TankError, e:
    do stuff
```

now it is 

```
try:
    download_path = sgtk.util.download_url
except TypeError:
    sgtk.util.download_url(self._bundle.shotgun, url, file_path)
except TankError, e:
    do stuff
```

Shouldn't it be nested, like this? 

```
try:

   try:
        download_path = sgtk.util.download_url
    except TypeError:
        sgtk.util.download_url(self._bundle.shotgun, url, file_path)

except TankError, e:
    do stuff
```

In the current implementation, what happens if `sgtk.util.download_url(self._bundle.shotgun, url, file_path)` raises a TankError?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -596,29 +680,51 @@ def _download_url(self, file_path, url, entity_type, entity_id, field):
...
+                    try:
+                        download_path = sgtk.util.download_url(
+                            self._bundle.shotgun, url, file_path, True
+                        )
+                        file_path = download_path
+                    except TypeError:
+                        # Probably an older version of core that doesn't have the
+                        # sgtk.util.download_url() `use_url_exension` arg implemented.
+                        # Try running without that arg value.
+                        sgtk.util.download_url(self._bundle.shotgun, url, file_path)
```
manneohrstrom: for the legacy fallbacks, do we want to append a hard coded jpg file extension? Or is that bad? Wasn't sure.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -679,22 +787,35 @@ def _get_thumbnail_path(url, bundle, directory_only=False):
...
+        cache_path_items = [
+            bundle.cache_location, "thumbs", first_folder, second_folder
+        ]
 
         # If we were only asked to give back a directory path then we can
         # skip building and appending a file name.
         if not directory_only:
-            path_chunks.append("%s.jpeg" % hash_str[4:])
+            path_base = hash_str[4:]
+            url_base, url_ext = os.path.splitext(url_obj.path)
```
manneohrstrom: please add a comments to help readers understand what these to returned chunks would typically contain

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -679,22 +787,35 @@ def _get_thumbnail_path(url, bundle, directory_only=False):
...
+            if url_ext:
+                cache_path_items.append("%s%s" % (path_base, url_ext))
```
manneohrstrom: wondering if we are are making this too complicated.

maybe it's simpler like this:

```
# Case A: Cache file does not exist

(path, cache_exists) = _get_thumbnail_path("https://foo/bar/baz.jpg") 
# return data (path, cache_exists) = ('/tmp/xx/yy/1245/6678', False) 
# we always get a path without extension back whenever a cache item doesn't exist
# we return False to indicate that a cache file cannot be found
# note: we don't try to deduce any file extension from the url 

# now we therefore download the thumb
full_path = sgtk.util.download_url("https://foo/bar/baz.jpg", path)
# download_url appends the extension and 
# full_path now contains /tmp/xx/yy/1245/6678.jpg

# Case B: cache is warm
(path, cache_exists) = _get_thumbnail_path("https://foo/bar/baz.jpg") 
# return data (path, cache_exists) = ('/tmp/xx/yy/1245/6678.jpg', True)
# all good job done
```

The cache file `/tmp/xx/yy/1245/6678` is unique and generated from the full url - we assume that 
there is only ever one `/tmp/xx/yy/1245/6678.*` file on disk - there cannot be say a `/tmp/xx/yy/1245/6678.jpg` and a `/tmp/xx/yy/1245/6678.png` - all we need to look for is the first match.

This behaviour seems simpler to me. What do you think? 

Also note my comments on the core pull request related to this - i think the download-url method should always **append** the extension just to keep things simple.

manneohrstrom: Adding this to the doc string with clear examples would be great too.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -162,16 +169,32 @@ def download_thumbnail(url, bundle):
...
+                    bundle.shotgun, url, path_to_cached_thumb, True
+                )
+                path_to_cached_thumb = full_path
+            except TypeError:
+                # This may be raised if an older version of core is in use
+                # that doesn't have the final `use_url_extension` arg implemented
+                # in sgtk.util.download_url() (set to True above). Since the url
+                # is not being checked for an extension, also revert to the
+                # previous behavior of _get_thumbnail_path() which hard-coded a
+                # ".jpeg" extension to the thumbnail file path.
```
manneohrstrom: great comment!

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -679,23 +867,44 @@ def _get_thumbnail_path(url, bundle, directory_only=False):
...
-        # join up the path
+            # Look for an existing cache file. Use the glob module since
+            # we do not know what the file type of the cache file is.
+            path_base = hash_str[4:]
+            cache_base = os.path.join(*(cache_path_items + [path_base]))
+
+            # Attempt to match something that looks like:
+            #   /bundle_cache_location/thumbs/C1C2/C3C4/rest_of_hash.*
+            cache_matches = glob.glob("%s.*" % cache_base)
+            if len(cache_matches):
```
manneohrstrom: not sure if it's overkill, but maybe worth throwing in a warning if there is more than 1 match?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -653,7 +813,34 @@ def _get_attachment_path(attachment_entity, bundle):
     @staticmethod
     def _get_thumbnail_path(url, bundle, directory_only=False):
         """
-        Returns the location on disk suitable for a thumbnail given its url.
+        Returns the location on disk suitable for a thumbnail given its url and
+        whether a cached file for the specified ``url`` already exists. Two cases
+        are handled:
+
+        Case A: ``directory_only`` is set to False and the ``url`` cache file does not exist:
```
manneohrstrom: this is unrelated to the CR but what's this directory_only business? It doesn't seem safe - as far as i can tell from the code, it's stripping out a bunch of the hashing that is supposed to guarantee that the cache is unique. Can you take a quick look? It looks like it's used for the attachments download - @thebeeland, what's the reason it needs to be cached differently? Could it be done without dropping parts of the hash? 

As far as i can tell, it looks like as if all attachment for example were named foo.png there would be a high risk of hash collisions if we only use the first 4 chars of the md5. 

If the attachment filename is important, perhaps we can append it and make it part of the filename or something like that?

PR #60, opened 2017-02-03
40080 download full thumbnail

akhronek:


python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -891,17 +891,26 @@ def _get_thumbnail_path(url, bundle, directory_only=False):
             #   /bundle_cache_location/thumbs/C1C2/C3C4/rest_of_hash.*
             cache_matches = glob.glob("%s.*" % cache_base)
             if len(cache_matches):
+                if len(cache_matches) > 1:
```
manneohrstrom: If i am reading this correctly, it seems like we are removing a `*.jpeg` file only to replace it with exactly the same `*.jpg` file? So this change in the shotgunutils framework will basically mean that all thumbnail cache are invalid and will be re-downloaded again. This doesn't seem great to me - why cannot this change be backwards compatible?

I thought we wanted to add the ability to download and cache `png` and other formats alongside the existing `jpeg` support. Unless we have strong reasons not to (and in that case let's discuss) the behaviour for caching `jpeg` files should remain unchanged.

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -891,17 +891,26 @@ def _get_thumbnail_path(url, bundle, directory_only=False):
...
+                    # There may be a legacy ".jpeg" file cached here. If so, remove it.
+                    old_cache_file = "%s.jpeg" % cache_base
+                    if old_cache_file in cache_matches:
+                        bundle.log_warning("Removing legacy '.jpeg' cache file : %s" % old_cache_file)
+                        cache_matches.remove(old_cache_file)
+                        os.remove(old_cache_file)
+
+                    # Now if more than one cache match is found, log a warning.
+                    if len(cache_matches) > 1:
+                        bundle.log_warning("More than one cached file found for url '%s':" % url)
```
manneohrstrom: This is a cache, so i don't think this is something the user needs to worry about. Warning messages are bubbled up to the user. When we report a warning, we basically are trying to tell the user "hey there is something we want you to be aware of - it's not necessarily wrong or an error, but you should be aware". This caching policy thing doesn't seem to fit in that category. I think it should be a debug. Does that make sense?

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -891,17 +891,26 @@ def _get_thumbnail_path(url, bundle, directory_only=False):
...
+                        [bundle.log_warning("    %s" % cache_match) for cache_match in cache_matches]
+                        bundle.log_warning(
+                            "Using '%s'.  If this is incorrect, try clearing your local site cache." %
```
manneohrstrom: If i am animator using zero config, this message will be very hard to understand! Again, all of these messages should probably be debug.

PR #65, opened 2017-02-28
Doc tweaks for hierarchy model limitations.

josh-t:
A client attempted to use the new API with an unsupported seed. This cleans up the docs to make the limitations clearer. Also hides the `NAV_ENTRIES` additional filter type from the docs to prevent direct usage until it is officially public.

### DIFF
--- a/python/shotgun_model/shotgun_hierarchy_item.py
+++ b/python/shotgun_model/shotgun_hierarchy_item.py
@@ -122,16 +122,16 @@ class ShotgunHierarchyItem(ShotgunStandardItem):
         The dictionary also stores a ``type`` key whose value is the type of
         entity being targeted.
 
-        A common usage of this is to respond to selection changed signals
-        on a view showing a shotgun hierarchy. For example, the data returned
-        can be fed to a separate shotgun model view.
+        A common usage of this is to respond to selection-changed signals
+        on a view showing a Shotgun hierarchy. For example, the data returned
+        can be fed to a separate Shotgun model view.
 
         Example usage::
 
             target_entities = selected_item.target_entities()
 
             sg_model.load_data(
-                target_entities.get("type"),
+                target_entities["type"],
                 additional_filter_presets=target_entities.get("additional_filter_presets"),
             )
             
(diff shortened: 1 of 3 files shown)


python/shotgun_model/shotgun_hierarchy_item.py
```diff
@@ -122,21 +122,19 @@ def target_entities(self):
...
-                        'preset_name': 'NAV_ENTRIES',
-                        'seed': {
-                            'field': 'entity',
-                            'type': 'Version'
-                        }
-                    }
-                ],
-                'type': 'Version'
-            },
+        A common usage of this is to respond to selection changed signals
```
thebeeland: selection-changed signals

python/shotgun_model/shotgun_hierarchy_item.py
```diff
@@ -122,21 +122,19 @@ def target_entities(self):
...
+        on a view showing a shotgun hierarchy. For example, the data returned
```
thebeeland: Capitalize "Shotgun"

thebeeland: Just below here, too.

python/shotgun_model/shotgun_hierarchy_item.py
```diff
@@ -122,21 +122,19 @@ def target_entities(self):
...
+        can be fed to a separate shotgun model view.
+
+        Example usage::
+
+            target_entities = selected_item.target_entities()
+
+            sg_model.load_data(
+                target_entities.get("type"),
```
thebeeland: [nitpicky] The .get is overkill here, since all entity dicts will contain a type.

josh-t: fixed. i copied form some source which should probably be changed as well. :P

python/shotgun_model/shotgun_hierarchy_item.py
josh-t: fixed both.

python/shotgun_model/shotgun_hierarchy_item.py
josh-t: changed.

PR #68, opened 2017-04-26
For #41930, loader hierarchy adjustment

jfboismenu:
his introduces a set of enhancements to the ShotgunHierarchyModel.

### Adds support for asynchronously resolving a series of nodes.
This allows to write UI code that will be able to populate the model with nodes from the root down to a specific one and then select that node in the tree.

### Root of the model is not a path
This allows to specify a Shotgun entity instead of passing in a string to initialize the model and delegates the responsibility of figuring out the path to the entity to the model.

I've also introduced some noise in the PR due to moving from log_debug to logger.debug. Sorry about that!

screen shot 2017-04-26 at 08 44 53

### DIFF
--- a/python/shotgun_model/shotgun_hierarchy_model.py
+++ b/python/shotgun_model/shotgun_hierarchy_model.py
@@ -472,8 +472,6 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
         # todo: hierarchy model to handle multiple rows?
         parent.appendRow(item)
 
-        parent.setData(not item.is_empty_kind(), self._SG_ITEM_HAS_CHILDREN)
-
         self._node_refreshed.emit(item)
 
     def _update_item(self, item, data_item):
--- a/python/shotgun_model/shotgun_hierarchy_model.py
+++ b/python/shotgun_model/shotgun_hierarchy_model.py
@@ -141,7 +141,7 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
             # our node is refreshed.
             model._node_refreshed.connect(self._node_refreshed)
 
-            self.parent()._log_debug("Fetching more on %s" % path_to_refresh[0])
+            print("Fetching more on %s" % path_to_refresh[0])
             # Fetch data from this path's parent.
             model.fetchMore(
                 model.item_from_path(path_to_refresh[0]).index()
@@ -151,8 +151,9 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
 
         def _node_refreshed(self, item):
             if item.data(self.parent()._SG_ITEM_UNIQUE_ID) != self._path_to_refresh[1]:
+                print "skipping node", item.data(self.parent()._SG_ITEM_UNIQUE_ID)
                 return
-            self.parent()._log_debug(
+            print(
                 "Model item refreshed: %s" % item.data(self.parent()._SG_ITEM_UNIQUE_ID)
             )
             self.parent()._node_refreshed.disconnect(self._node_refreshed)
@@ -169,7 +170,7 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
         if not paths:
             return
 
-        self._log_debug("Async loading of %s" % paths)
+        print("Async loading of %s" % paths)
 
         for idx, path in enumerate(paths):
             # Iterate on every path.
@@ -182,11 +183,12 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
             # This time around this node will already have been refreshed
             # and the code will dig deeper. At some point the last entry
             # in the list will be reached and we will emit the item.
-            self._log_debug("Refreshing paths: %s" % paths)
+            print("Refreshing paths: %s" % paths[idx - 1:])
             self._NodeRefresher(paths[idx - 1:], self)
             return
 
         print("Deep load has been completed for %s" % paths[-1])
+        print("Here's the winner: %s" % item)
         # If everything is loaded, emit the signal.
         self.deep_load_completed.emit(item)
 
@@ -460,6 +462,8 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
         # todo: hierarchy model to handle multiple rows?
         parent.appendRow(item)
 
+        self._node_refreshed.emit(item)
+
     def _update_item(self, item, data_item):
         """
         Updates a model item with the given data
@@ -492,7 +496,11 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
         # run the finalizer (always runs on construction, even via cache)
         self._finalize_item(item)
 
-        self._node_refreshed.emit(item)
+        # If this is a node that already existed and is being refreshed, notify.
+        # If the node is not parented yet, this is because we're in the middle
+        # of a create_item call, which WILL refresh the node.
+        if item.parent():
+            self._node_refreshed.emit(item)
 
     ############################################################################
     # private methods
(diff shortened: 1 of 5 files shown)


python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -60,6 +70,8 @@ def __init__(self, parent, schema_generation=0, bg_task_manager=None):
         """
         super(ShotgunHierarchyModel, self).__init__(parent, bg_task_manager)
 
+        self._logger = sgtk.platform.get_logger(__name__)
```
manneohrstrom: our recommended pattern is to keep a logger global at top of the file, not per class instance like this. I suggest moving it to the top unless you have a good reason not to (in that case, please add a comment!)

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -119,8 +131,141 @@ def item_from_path(self, path):
             :meth:`~shotgun-api3:shotgun_api3.Shotgun.nav_expand()` method.
         :returns: :class:`~PySide.QtGui.QStandardItem` or ``None`` if not found
         """
-        self._log_debug("Resolving model item for path %s" % path)
-        return self._ensure_item_loaded(path)
+        self._logger.debug("Resolving model item for path %s" % path)
```
manneohrstrom: just `logger.debug("xxx")` once you have moved it to the top of the file. That's the recommended pattern these days.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -119,8 +131,141 @@ def item_from_path(self, path):
...
+
+        if path == self._path:
+            return self.invisibleRootItem()
```
manneohrstrom: cleaner with a single return at the end of the method?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -119,8 +131,141 @@ def item_from_path(self, path):
...
+
+            :param list(str) path_to_refresh: List of nodes to refresh asynchronously.
+            :param model: ``ShotgunHierarchyModel`` we are requesting the nodes for.
+            """
+            super(ShotgunHierarchyModel._NodeRefresher, self).__init__(model)
+            # Connect to the node refreshed signal so we know when
+            # our node is refreshed.
+            model._node_refreshed.connect(self._node_refreshed)
+
+            self.parent()._logger.debug("Fetching more on %s" % path_to_refresh[0])
```
manneohrstrom: just `logger.debug()` will do once you have moved logging up.

python/shotgun_model/shotgun_hierarchy_model.py
jfboismenu: Yup, I was working around a limitation of our doc generating script, but I've now fixed it, so we're all good.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -119,8 +134,141 @@ def item_from_path(self, path):
...
-        self._log_debug("Resolving model item for path %s" % path)
-        return self._ensure_item_loaded(path)
+        logger.debug("Resolving model item for path %s" % path)
+
+        if path == self._path:
+            return self.invisibleRootItem()
+        else:
+            return self._ensure_item_loaded(path)
+
+    class _NodeRefresher(QtCore.QObject):
```
josh-t: Does this need to be contained in the Hierarchy model?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -119,8 +134,141 @@ def item_from_path(self, path):
...
+            # If an item is already loaded, move to the next one.
+            if item:
+                continue
+            # Send a refresh request and return. When the node is refreshed,
+            # this method will be called a second time with the same parameters.
+            # This time around this node will already have been refreshed
+            # and the code will dig deeper. At some point the last entry
+            # in the list will be reached and we will emit the item.
+            logger.debug("Refreshing paths: %s", paths[idx - 1:])
+            self._NodeRefresher(paths[idx - 1:], self)
```
josh-t: Any concerns about not keeping references to these objects as they're created?

python/shotgun_model/simple_shotgun_hierarchy_model.py
```diff
@@ -31,7 +31,7 @@ class SimpleShotgunHierarchyModel(ShotgunHierarchyModel):
     can also be subclassed from this class.
     """
 
-    def load_data(self, seed_entity_field, path=None, entity_fields=None):
+    def load_data(self, seed_entity_field, root=None, entity_fields=None):
```
josh-t: This is a breaking change to a public API. Should we add some backward compatibility?

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -60,6 +73,8 @@ def __init__(self, parent, schema_generation=0, bg_task_manager=None):
         """
         super(ShotgunHierarchyModel, self).__init__(parent, bg_task_manager)
 
+        logger = sgtk.platform.get_logger(__name__)
```
jfboismenu: Already done!

python/shotgun_model/shotgun_hierarchy_model.py
jfboismenu: I didn't want to polute the global scope.

josh-t: For me, classes within classes just makes the code more difficult to read. My opinion is that unless we have a good reason, we should keep them separate. I'm not super passionate about it though.

python/shotgun_model/shotgun_hierarchy_model.py
jfboismenu: Actually, the constructor of _NodeRefresher does. The refresher is a [child of the model](https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/develop/zero_config/python/shotgun_model/shotgun_hierarchy_model.py#L155). And we delete later when we're done.!

josh-t: cool!

python/shotgun_model/simple_shotgun_hierarchy_model.py
jfboismenu: @manneohrstrom said we should allow to break this one since it uses an API that is internal (nav_api) and as such is experimental. Do you still think so Manne?

josh-t: That's cool. Perhaps we should add back a `path=None` and have a more direct `raise` with a deprecation error so that it's clear to people when they try to use `path`? We could add a ticket to remove it after some time as well. Not a huge deal, but maybe better than the standard error saying unrecognized argument or whatever.

python/shotgun_model/shotgun_hierarchy_model.py
manneohrstrom: for clarify maybe we could do `self._refreshed= self._NodeRefresher` ? the model child relationship is an internal C++ binding only, always good to keep a python ref too for good measure IMO :)

python/shotgun_model/simple_shotgun_hierarchy_model.py
manneohrstrom: @josh-t your call, you know this better than i do. I think either we do 100% backwards compatibility or drop it completely. no middle ground solution.

josh-t: No worries. I was just being cautious. All good to push forward with JF's changes. :)

PR #69, opened 2017-05-03
41668 Adds ability to include root item as top-level sibling in hierarchy model

josh-t:
This change facilitates UIs that need to display target entities associated with the root of a hierarchy model. You can now supply an `include_root` string to the hierarchy constructor which indicates that the root item should be included as a sibling to the top-level children. While this is a but unorthodox, it does make for an easy way to select the root item in a view attached to the model which may trigger display of the associated entities. An example of this use case is the loader hierarchy and the need to display published file entities attached to the project. Prior to this change, when setting the project entity as the root, there was no way to access the project for selection and therefore no way to access the project-level publishes.

### DIFF
--- a/python/shotgun_model/data_handler_nav.py
+++ b/python/shotgun_model/data_handler_nav.py
@@ -8,6 +8,7 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
+import copy
 import gc
 
 from .data_handler import ShotgunDataHandler, log_timing
@@ -31,7 +32,7 @@ class ShotgunNavDataHandler(ShotgunDataHandler):
     _SG_PATH_FIELD = "path"
     _SG_PARENT_PATH_FIELD = "parent_path"
 
-    def __init__(self, root_path, seed_entity_field, entity_fields, cache_path):
+    def __init__(self, root_path, seed_entity_field, entity_fields, cache_path, include_root=None):
         """
         :param str root_path: The path to the root of the hierarchy to display.
             This corresponds to the ``path`` argument of the
@@ -52,11 +53,22 @@ class ShotgunNavDataHandler(ShotgunDataHandler):
             of field names to return.
 
         :param str cache_path: Path to cache file location.
+
+        :param str include_root: Defines the name of an additional, top-level
+            model item that represents the root. In views, this item will appear
+            as a sibling to top-level children of the root. This allows for
+            UX whereby a user can select an item representing the root without
+            having a UI that shows a single, top-level item. An example would
+            be displaying published file entity hierarchy with top level items:
+            "Assets", "Shots", and "Project Publishes". In this example, the
+            supplied arg would look like: ``include_root="Project Publishes"``.
+            If ``include_root`` is ``None``, no root item will be added.
         """
         super(ShotgunNavDataHandler, self).__init__(cache_path)
         self.__root_path = root_path
         self.__seed_entity_field = seed_entity_field
         self.__entity_fields = entity_fields
+        self.__include_root = include_root
 
     def generate_data_request(self, data_retriever, path):
         """
@@ -132,12 +144,6 @@ class ShotgunNavDataHandler(ShotgunDataHandler):
 
         self._log_debug("Generating new tree in memory...")
 
-        if item_path == self.__root_path:
-            self._log_debug("This is the root of the tree.")
-            parent_uid = None
-        else:
-            parent_uid = item_path
-
         # create a brand new tree rather than trying to be clever
         # about how we cull intermediate nodes for deleted items
         diff_list = []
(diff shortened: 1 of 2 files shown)


python/shotgun_model/data_handler_nav.py
```diff
@@ -146,10 +152,40 @@ class and the data payload from Shotgun should be provided via the
...
         new_uids = set()
+
+        # a list of sg data dicts to display items for
+        child_data = []
+
+        if item_path == self.__root_path:
+            self._log_debug("This is the root of the tree.")
+            parent_uid = None
+
+            if self.__include_root:
```
jfboismenu: I'm really worried this will break something. Did you try the search box and expand an asset? How about the home button? Does it take you to the project in the hierarchy?

There's this [bit of code](https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/develop/zero_config/python/shotgun_model/shotgun_hierarchy_model.py#L124-L140) in the model that assume the path `/Project/<id>` is at the root of model. By having an extra node with the same path somewhere in the tree we might be looking for trouble.

josh-t: Totally on board with your concerns. Search box, asset expansion, home button all appear to be working fine. The home button takes you to the current context. If the context is the project, it takes you to the unselected state which is what I expect. 

Regarding the path, I was originally removing it all together from the special item and it was working fine. Felt a bit unnecessary, but maybe it should be that way.

josh-t:
Understand the concerns. It's really not that much of a hack I don't think. It's just adding an additional item to the model, as called out by the client code. It's not creating it out of thin air, it's still driven by the nav api, only duplicating the root and making it visible. It definitely feels weird, so I agree there's probably another way to do this, but I'm concerned about over engineering this solution if this works and doesn't break anything. Agree that @manneohrstrom should weigh in. :)

python/shotgun_model/data_handler_nav.py
jfboismenu: The home button is supposed to take you some place that makes sense for your current context.

- If you are on a entity or task context, it should take you to an entity.
- If you a project context, it should take you to the project

I would have expected the Home button to take me to the project with your changes. The reason why it might not have done what I expected is because I didn't write the code correctly originally. :)

On [this line](https://github.com/shotgunsoftware/tk-multi-loader2/blob/develop/zero_config/python/tk_multi_loader/dialog.py#L828), I implemented the home button behaviour for the hierarchy.

It should probably have read something like `ctx.entity or ctx.project`. What happens if you pick that?

josh-t: I'd suggest that the unselected, default view is the best view for the project context because it shows you the child folders on the right. If it selected the special item, it would only show top-level publishes (not the subfolders). It feels right to me as is. Making the change you suggest doesn't seem to change the behavior.

manneohrstrom:
Note: zero config branch has been merged so i shifted the PR target branch to be master instead. master should be identical to zero config, so not sure why there are conflicts reported. Ping me if you have questions!

python/shotgun_model/data_handler_nav.py
manneohrstrom: i agree with @jfboismenu that prior to having this special node, clicking the home button when it's linked to the project context should show the treeview (say in the loader) unselected and with all children visible - but with this change it feels like the new special root path should be selected.

manneohrstrom:
note: `develop/zero_config` has been merge. Pls merge into master.

python/shotgun_model/data_handler_nav.py
```diff
@@ -52,11 +53,22 @@ def __init__(self, root_path, seed_entity_field, entity_fields, cache_path):
...
+            be displaying published file entity hierarchy with top level items:
+            "Assets", "Shots", and "Project Publishes". In this example, the
+            supplied arg would look like: ``include_root="Project Publishes"``.
+            If ``include_root`` is ``None``, no root item will be added.
         """
         super(ShotgunNavDataHandler, self).__init__(cache_path)
         self.__root_path = root_path
         self.__seed_entity_field = seed_entity_field
         self.__entity_fields = entity_fields
+        self.__include_root = include_root
```
josh-t: I'm fine to use the item label for display, but I think it will render this solution less effective. Again, I know this is an odd solution, but using the root entity label will add confusion I think and it will also mean we probably need to do additional work to make it show up on top.

python/shotgun_model/data_handler_nav.py
josh-t: I think I'm going to have to disagree with you guys on this one. Remember, the special item isn't the entire project view of the hierarchy. It's a special item that is only showing target entities attached at that level (no subfolders). The deselected state is the full project context and I think that's right still. Maybe try it out on the loader branch? I think it might be clearer once you've interacted with it? I'm willing to concede if you try it out and still feel strongly that it should select the special item.

josh-t:
@manneohrstrom regarding the deselected state showing publishes next to the top-level folders (children). This was my original plan, but there are a couple of things that I think make this non-trivial.

1. We'd have to alter the selection mode or add the ability to deselect items. Probably not a huge hurdle, but it made me nervous thinking about it. 
2. We don't currently store information about the root item returned from `nav_expand()` we basically throw it away, build items for the children, and attach them to the invisible root item. That's why it worked when I forced the path to always be "/" to display all projects. Since there was an item in the tree for the project, it could be selected which would display the children. Same reason the "special" item works. 

Having typed all that, we could try another approach of building the hierarchy with a single root item and using `setRootIndex()` on the view to display only the children. This would make the project item invisible, but we would have the data in the model. We're still left with the deselect problem, but I'm looking on stack overflow and it may not be horrible to address. I'll take a crack at it tomorrow. Stay tuned...

manneohrstrom:
we now have a solution for this so i am closing the PR.

PR #70, opened 2017-05-03
#42432 Hierarchy search and expansion utils

manneohrstrom:
- Adds data retriever support for nav search methods
- Adds async expansion/fetch_more methods to the hierarchy model via the async_item_from_entity() and async_item_from_paths () methods.
- upgrades to 0.18 logging
- the hierarchy model _load_data method has a breaking interface change (note: it's not officially released yet). Previously, it would take a path argument, now it takes a root argument instead in order to avoid using path strings to drive the API.


manneohrstrom:
----------
This branch has been merged but is left open to ensure the ß can update itself properly.
DO NOT COMMIT ANY FURTHER CHANGES TO THIS BRANCH.

manneohrstrom:
beta is over. deleting branch.

PR #72, opened 2017-08-17
For #40555: Only uses deleteLater if we're in Qt4.

thebeeland:
We're seeing a lot of problems related to garbage collection in PySide2/Qt5 when deleteLater is used. These issues result in crashes on DCC close in some cases, but also errors on app close even when the DCC isn't being closed.

### DIFF
--- a/python/shotgun_model/shotgun_hierarchy_model.py
+++ b/python/shotgun_model/shotgun_hierarchy_model.py
@@ -22,9 +22,8 @@ from .shotgun_hierarchy_item import ShotgunHierarchyItem
 from .shotgun_query_model import ShotgunQueryModel
 from .data_handler_nav import ShotgunNavDataHandler
 from .util import sanitize_for_qt_model
-from ..utils import safe_delete_later
 
-
+utils = sgtk.platform.current_bundle().import_module("utils")
 logger = sgtk.platform.get_logger(__name__)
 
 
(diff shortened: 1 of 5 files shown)


python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -203,7 +203,13 @@ def _node_refreshed(self, item):
                 "Model item refreshed: %s", item.data(self.parent()._SG_ITEM_UNIQUE_ID)
             )
             self.parent()._node_refreshed.disconnect(self._node_refreshed)
-            self.deleteLater()
+            # We've seen problems in Qt5 with deleteLater causing
```
josh-t: Same comment as the qtwidgets CR. I think if we're going to do this, we should do it via method in `sgtk.platform.qt`. What do you think?

python/task_manager/worker_thread.py
```diff
@@ -67,44 +67,56 @@ def run(self):
...
+                    self._mutex.lock()
+                    try:
+                        if not self._process_tasks:
+                            break
+                        tb = traceback.format_exc()
+                        # emit failed signal (non-blocking):
+                        self._results_dispatcher.emit_failure(self, task_to_process, str(e), tb)
+                    finally:
+                        self._mutex.unlock()
+        except RuntimeError:
```
josh-t: Would it make sense to manually call `shut_down` in this case?

thebeeland: It's likely that shut_down has already been called at this point. It appears to be a race condition of some kind that I was battling, and this was the only way to suppress it reliably.

python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -22,6 +22,7 @@
 from .shotgun_query_model import ShotgunQueryModel
 from .data_handler_nav import ShotgunNavDataHandler
 from .util import sanitize_for_qt_model
+from ..utils import safe_delete_later
```
josh-t: You may want to try to build the sphinx docs to make sure this works. I think you may have to do the toolkit-style import:

```python
utils = sgtk.platform.current_bundle().import_module("utils")
```

PR #74, opened 2017-11-23
#45562 fix for model population on new items added

pscadding:
This fixes a bug where, it would only display new items in the treeview, if the cached data hadn't already been added to the model.

To Repro, 
- Have a couple of assets of the same asset type on your project
- Set your workfiles app to have `auto_expand_tree: false`
- load workfiles, expand asset type node to see your assets
- close workfiles and add a new asset in shotgun with the same type
- open workfiles, expand asset type node, you will only see the new asset, until you reload the app.

The issue happens because it only loads the cache data if there are no child items already added to the model:
https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/dd5058426e32d02cd6478e95f2fb0954498b8d1c/python/shotgun_model/shotgun_query_model.py#L375

If you have `autoexpand` enabled, then it will have already loaded the cached items prior to finding any new entities, so in that scenario, everything is fine, as the cached items are already added, and the new items get added to the model in an ad-hoc way. However if the nodes are not set to expand, it defers the adding of the cached data until the node is expanded, but the new items have already been added by this point so the check decides that it doesn't need to add the cache data, hence the bug.

Ideally, the fix would have happened [here](https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/dd5058426e32d02cd6478e95f2fb0954498b8d1c/python/shotgun_model/shotgun_query_model.py#L375), and I experimented with checking if the model contained the same UIDs as the cached data, and if not, then I would pass back true to enable it to run the fetchMore method. However as the new items had already been added to the model but also already added to the cache, this resulted in duplicates for the new items. I would have had to add a check in probably somewhere [here](https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/dd5058426e32d02cd6478e95f2fb0954498b8d1c/python/shotgun_model/shotgun_model.py#L579-L604) to check for duplicates, but this didn't feel like a good fix. 
So instead I opted to check at the point where the new items were detected and added, and only add them if the cache data had already been added. 
This seems to work, but my main concern is my use of the line:
[`pre_update_cache = copy.copy(self._data_handler._cache)`](https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/45562_fix_for_model_population/python/shotgun_model/shotgun_query_model.py#L964)
becuase I am accessing the private variable `_cache`, and I'm also copying the data so that I can compare the cache before it got the new items added.

### DIFF
--- a/python/shotgun_model/shotgun_query_model.py
+++ b/python/shotgun_model/shotgun_query_model.py
@@ -11,6 +11,7 @@
 import sgtk
 import weakref
 import datetime
+import copy
 
 # NOTE: This is a dummy call to work around a known bug in datetime
 # whereby there is code imported at call time that is done so in a
@@ -957,6 +958,11 @@ class ShotgunQueryModel(QtGui.QStandardItemModel):
         # pre-process data
         sg_data = self._before_data_processing(sg_data)
 
+        # take a copy of the data cache before we add any new items to it.
+        # this is so we can compare the pre updated cache against the current items in the model
+        # So we can work out if we want add the new items now, or wait until the fetchMore adds all the cached items
+        pre_update_cache = copy.copy(self._data_handler._cache)
+
         # push shotgun data into our data handler which will figure out
         # if there are any changes
         self._log_debug("Updating data model with new shotgun data...")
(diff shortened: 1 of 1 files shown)


python/shotgun_model/shotgun_query_model.py
```diff
@@ -957,6 +958,11 @@ def __on_sg_data_arrived(self, sg_data):
         # pre-process data
         sg_data = self._before_data_processing(sg_data)
 
+        # take a copy of the data cache before we add any new items to it.
+        # this is so we can compare the pre updated cache against the current items in the model
+        # So we can work out if we want add the new items now, or wait until the fetchMore adds all the cached items
+        pre_update_cache = copy.copy(self._data_handler._cache)
```
josh-t: I agree with your comment that this is not ideal (accessing a protected member of the data handler). Maybe we could add a method to the data handler to call into the cache: `cached_child_uids = self._data_handler.get_child_uids(uid)`.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -1003,8 +1009,32 @@ def __on_sg_data_arrived(self, sg_data):
...
+                        # to add our new item to the cache and the let the fetchMore method add it to the model.
+
+                        uid = parent_model_item.data(self._SG_ITEM_UNIQUE_ID)
+
+                        # get a set of UIDs that are already added to the model
+                        current_child_items = set(
+                            [parent_model_item.child(row).data(self._SG_ITEM_UNIQUE_ID) for row in range(parent_model_item.rowCount())])
+
+                        # get a set of UIDs that existed in the cache before the new items were added
+                        cached_child_items = set(pre_update_cache.get_child_uids(uid))
```
josh-t: Can we do this outside the loop as suggested above?

manneohrstrom:
Fantastic work @pscadding and what a great PR description. @josh-t 's suggestion sounds great to me - i think it's worth exploring.

sgsteph:
Simpler solution in this PR #77

jfboismenu:
Closing at the fix was implemented and released from #77 .

PR #75, opened 2017-12-04
43294 cache sharing

sgsteph:
Allow to share data cache location for thumbnails and SG queries.

Check if the`Bundle.site_cache_location` which was just introduced in tk-core is available, leverage it if so, fallback on the caching per project/pipeline config/plugin id if not.


sgsteph:
Joint tk-core pull request is here: https://github.com/shotgunsoftware/tk-core/pull/560

PR #76, opened 2018-01-16
39289 old cached data culling

sgsteph:
These changes allow the new `Bundle._remove_old_cached_data` to be used for this framework.

Culling old data is based on modification times, so we ensure that needed data will not be removed when culling old data by updating the files modification time when they are re-used. This is needed for thumbnails, schema and status caches, and attachments. This is not needed for cached SG queries, as they are usually always refreshed.

A data clean up is posted when the framework is loaded in `tk-desktop`. I'm not quite sure this is the right place to do it, any better place is welcome. In any case, it shouldn't be triggered too often: we definitely don't want to clean up old data every single time the framework is loaded.

I added some unit tests. Unfortunately dealing with a background task manager in unit tests is far from being easy, so I took some shortcuts and didn't implement tests for schema and status caching. Please let me know if that looks ok.

tk-core pull request with the new method is here: https://github.com/shotgunsoftware/tk-core/pull/587


python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -203,7 +203,12 @@ def download_thumbnail(url, bundle):
                 os.chmod(path_to_cached_thumb, 0666)
             finally:
                 os.umask(old_umask)
-
+        else:
+            # Update access and modified time to "now" so the file will be kept
+            # around when culling old files in the cache.
+            # `_get_thumbnail_path` returns a full path with the extension if
+            # the thumb exists.
+            os.utime(path_to_cached_thumb, None)
```
manneohrstrom: maybe, for clarity, we could have this in a method that is being called? something like - `self._indicate_resource_accessed()` or something?

python/shotgun_globals/cached_schema.py
```diff
@@ -191,6 +191,9 @@ def _load_cached_status(self, project_id=None):
         if os.path.exists(status_cache_path):
             try:
                 self._bundle.log_debug("Loading cached status from '%s'" % status_cache_path)
+                # Reset the file modification time to prevent it to be removed
```
manneohrstrom: i think we should just clean up thumbnails and caches (which are an ever growing number of items) - the schema is just a constant number of files, so no need to clean up i think.

framework.py
```diff
@@ -17,8 +18,45 @@ class ShotgunUtilsFramework(sgtk.platform.Framework):
...
+            return
+
+        try:
+            self.log_info(
+                "Posting old cached data clean up for engine %s...." % self.engine.instance_name
+            )
+            # Qt might not be yet available at this stage (e.g. in tk-desktop),
+            # so use regular Python Thread to post the clean up in the background.
+            self._bg_cleanup_thread = threading.Thread(
+                target=self._remove_old_cached_data,
```
manneohrstrom: i am worried that we are over-thinking this. I think, at least for the time being, we should keep this out of core and inside this framework. These kind of dependencies can be a challenge to manage.

framework.py
```diff
@@ -17,8 +18,45 @@ class ShotgunUtilsFramework(sgtk.platform.Framework):
...
+        # This Bundle method was introduced in tk-core > v0.18.124
+        # Check the existence of the attribute to not introduce a dependency
+        # on tk-core release.
+        if not hasattr(self, "_remove_old_cached_data"):
+            return
+        # Cleaning up data can be slow if a lot of old data is present. As a
+        # safety measure, we only post it when running from tk-desktop, which
+        # allows us to not trigger the clean up every single time this framework
+        # is loaded.
+        if self.engine.instance_name != "tk-desktop":
```
manneohrstrom: this feels hacky. for example, it will break with desktop2. :)
I think we should always run it.

framework.py
```diff
@@ -17,8 +18,45 @@ class ShotgunUtilsFramework(sgtk.platform.Framework):
             
     def init_framework(self):
         self.log_debug("%s: Initializing..." % self)
+        self._bg_cleanup_thread = None
+        self._post_old_data_cleanup()
     
     def destroy_framework(self):
         self.log_debug("%s: Destroying..." % self)
+        if self._bg_cleanup_thread:
+            if self._bg_cleanup_thread.isAlive():
```
manneohrstrom: the worker thread is looping over files, deleting them. I think here, rather than waiting, we should just set a bool `stop_thread=True` causing the thread to exit the loop and complete, so that shutdown is instant. Then the next time the fw starts up, it will just continue.

sgsteph:
Ready for another review round, summary of what I changed:
* Moved `_remove_old_cached_data` from tk-core into the framework.
* Added a `_ALWAYS_KEEP_CACHED_FILES` class member to specify a list of files which should never be removed and removed the code updating the modification time for the files in this list (sg_status.pickle and sg_schema.pickle).
* Added a way to stop a clean up immediately, but we still have to join the thread, which should be immediate now.
* A cleanup is now always posted when the framework is loaded and I removed the hard coded tk-desktop case. It means that every single time the framework is loaded a cleanup will be posted and in some cases immediately stopped (e.g when launching a dcc with `./tank maya_2017`). I ran some tests, and it seems to behave, but please let me know if you have any worry about it.
* I added a `_indicate_resource_accessed` helper, it is a function and not a method because it needs to be used in static methods.
* I made clearer why we're using regular threads and not a bg manager or QThreads.
* I changed the time comparison test to only use timestamps without going through datetime conversions at it seems there are some float precision problems when doing this.

Let me know if it looks better!

python/shotgun_data/shotgun_data_retriever.py
```diff
@@ -18,6 +18,12 @@
 from sgtk.platform.qt import QtCore, QtGui
 from sgtk import TankError
 
+def _indicate_resource_accessed(file_path):
+    """
+    Helper to indicate a resource was accessed and shouldn't be considered
+    old when cleaning up old cached data.
+    """
+    os.utime(file_path, None)
```
manneohrstrom: nice!!!

framework.py
```diff
@@ -9,16 +9,150 @@
...
+import threading
+import datetime
+import os
+import time
 
 class ShotgunUtilsFramework(sgtk.platform.Framework):
     
+    # A list of file names which should never be deleted when cleaning up old
+    # cached data.
+    _ALWAYS_KEEP_CACHED_FILES = [
```
manneohrstrom: i know that this is cache data, but even so i am a bit paranoid about deleting. I would prefer if we had an opt-in approach rather than opt out. Instead of saying "delete everything except the schema files" we should say "delete thumbs and caches" - to avoid deleting something unintentionally. If we add some module to the sg utils in the future, it's easy to forget to add it to exclude list and we won't notice the behaviour until much later and the deletion kicks in. Also, maybe i am overthinking but there could be forks of sgutils where clients cache data etc. Just seems safer to be explicit in this case.

framework.py
```diff
@@ -9,16 +9,150 @@
...
     def destroy_framework(self):
+        """
+        Destroy this framework.
+
+        If an old cached data cleanup was posted in the background, stop it
+        immediately.
+        """
         self.log_debug("%s: Destroying..." % self)
+        # Please note that we are modifying a member which is read in another
+        # thread which should be fine in Python with the GIL protecting its access.
```
manneohrstrom: nice comment!

framework.py
```diff
@@ -9,16 +9,150 @@
...
+        self._stop_cleanup = True
+        if self._bg_cleanup_thread:
+            if self._bg_cleanup_thread.isAlive():
+                # If the clean up is not completed yet, log why we are waiting.
+                self.log_info("Waiting for old data clean up to complete...")
+            self._bg_cleanup_thread.join()
```
manneohrstrom: nitpick: do we need to wait for the thread if it is not alive? should this line be inside the if clause?

framework.py
```diff
@@ -9,16 +9,150 @@
...
+            # regular Python Thread to post the clean up in the background.
+            self._bg_cleanup_thread = threading.Thread(
+                target=self._remove_old_cached_data,
+                name="%s Clean Up" % self.name
+            )
+            self._bg_cleanup_thread.start()
+        except Exception as e:
+            self.log_warning("Unable to post data clean up: %s" % e)
+
+    def _remove_old_cached_data(self, grace_period=7):
```
manneohrstrom: Would be good if we defined the grace_period as a big constant that belongs to the class. And set it to be 60 days.

framework.py
```diff
@@ -9,16 +9,150 @@
...
+
+        :param int grace_period: The number of days files without any modification
+                                 should be kept around before being deleted.
+        :raises: ValueError if the grace_period is smaller than 1.
+        """
+        if grace_period < 1:
+            raise ValueError(
+                "Invalid grace period value %d, it must be a least 1" % grace_period
+            )
+        self.logger.debug("Starting old data cleanup...")
```
manneohrstrom: please include the grace period date in this message - that seems useful, maybe also explain what it will do. Something like "Cleaning all files with a modofication date older than xxxx in location ~/Library/caches/.../thumbnails"

framework.py
sgsteph: I think it is good practice to always join threads (or processes). The `if` is here only to log something if we think `join` will not return immediately. With the _stop cleanup_ hand break I added following your comments, I never saw the log message during my tests.

framework.py
sgsteph:
```
Cleaning all files with a modification date older than 2017-12-18
18:32:16.358171 under locations
/Users/steph/Library/Caches/Shotgun/couchlab/site/fw-shotgunutils/sg,
/Users/steph/Library/Caches/Shotgun/couchlab/site/fw-shotgunutils/sg_nav,
/Users/steph/Library/Caches/Shotgun/couchlab/site/fw-shotgunutils/thumbs,
/Users/steph/Library/Caches/Shotgun/couchlab/p309c171/fw-shotgunutils/sg,
/Users/steph/Library/Caches/Shotgun/couchlab/p309c171/fw-shotgunutils/sg_nav,
/Users/steph/Library/Caches/Shotgun/couchlab/p309c171/fw-shotgunutils/thumbs
```

PR #81, opened 2018-02-12
For #45552, broken QWaitCondition on CentOS 7

jfboismenu:
Swaps out the usage of QWaitCondition and QMutex in our code since CentOS 7 has issues with those when using PySide 2.

### DIFF
--- a/python/task_manager/worker_thread.py
+++ b/python/task_manager/worker_thread.py
@@ -14,6 +14,7 @@ Worker thread for the background manager.
 
 import traceback
 from sgtk.platform.qt import QtCore
+from threading import Lock, Condition
 
 
 class WorkerThread(QtCore.QThread):
@@ -33,8 +34,8 @@ class WorkerThread(QtCore.QThread):
 
         self._task = None
         self._process_tasks = True
-        self._mutex = QtCore.QMutex()
-        self._wait_condition = QtCore.QWaitCondition()
+        self._mutex = Lock()
+        self._wait_condition = Condition(self._mutex)
         self._results_dispatcher = results_dispatcher
 
     def run_task(self, task):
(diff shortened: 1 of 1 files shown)


python/task_manager/worker_thread.py
```diff
@@ -43,24 +44,18 @@ def run_task(self, task):
...
         """
         self._results_dispatcher = None
-        self._mutex.lock()
-        try:
+        with self._mutex:
             self._process_tasks = False
-        finally:
-            self._mutex.unlock()
-        self._wait_condition.wakeAll()
+            self._wait_condition.notifyAll()
```
jfboismenu: With `threading.Condition`, you need to own the lock to be able to notify, which if fine here, as the lock is released when notifying so someone else can grab it.

python/task_manager/worker_thread.py
```diff
@@ -71,44 +66,35 @@ def run(self):
...
-                self._mutex.lock()
-                try:
+                with self._mutex:
                     while self._process_tasks and not task_to_process:
                         task_to_process = self._task
                         self._task = None
                         if not task_to_process:
                             # wait until we have something to do...
-                            self._wait_condition.wait(self._mutex)
+                            self._wait_condition.wait()
```
jfboismenu: Python takes the lock as a parameter in the `__init__` instead of the wait method.

python/task_manager/worker_thread.py
```diff
@@ -43,24 +44,18 @@ def run_task(self, task):
 
         :param task:    The task to run
         """
-        self._mutex.lock()
-        try:
+        with self._mutex:
```
jfboismenu: Thank god for the `with` statement. I don't like the `try:` `finally:` pattern.

PR #87, opened 2018-03-26
#46776 Websockets support

manneohrstrom:
Bug fixes and extensions to help support websockets using the external config interfaces:
- Support for grouping and group defaults
- Support for shotgun specific permission command hints
- Support for requesting a general set of commands for an entity type (e.g. 'give me actions for assets')
- Bug fixes.

### DIFF
--- a/python/external_config/external_command.py
+++ b/python/external_config/external_command.py
@@ -272,6 +272,13 @@ class ExternalCommand(object):
         return self._pipeline_config_name
 
     @property
+    def system_name(self):
+        """
+        The system name for the command
+        """
+        return self._callback_name
+
+    @property
     def display_name(self):
         """
         Display name, suitable for display in a menu.
--- a/python/external_config/external_command.py
+++ b/python/external_config/external_command.py
@@ -8,6 +8,7 @@
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
 import os
+import sys
 import cPickle
 import sgtk
 from sgtk.util.process import subprocess_check_output, \
@@ -15,8 +16,6 @@ from sgtk.util.process import subprocess_check_output, \
 
 logger = sgtk.platform.get_logger(__name__)
 
-# file format magic number
-CONFIGURATION_GENERATION = 2
 
 
 class ExternalCommand(object):
@@ -30,6 +29,9 @@ class ExternalCommand(object):
     A command is executed via its :meth:`execute` method, which
     will launch it in the given engine.
     """
+
+    # file format magic number
+    FORMAT_GENERATION = 3
 
     @classmethod
     def serialize_command(cls, entity_type, command_name, properties):
@@ -59,12 +61,43 @@ class ExternalCommand(object):
             "group_default": properties.get("group_default") or False,
 
             # special for shotgun
-            "deny_permissions": properties.get("deny_permissions"),
-            "deny_platforms": properties.get("deny_platforms"),
-            "supports_multiple_selection": properties.get("supports_multiple_selection"),
+            "sg_deny_permissions": properties.get("deny_permissions"),
+            "sg_supports_multiple_selection": properties.get("supports_multiple_selection"),
         }
 
         return data
+
+    @classmethod
+    def enabled_on_current_os(cls, properties):
+        """
+        Checks toolkit command properties to determine
+        if a command is enabled on the current os or Not.
+
+        :param dict properties: Properties dictionary
+            as returned by the ``Engine.commands`` dictionary property.
+        :return: True if enabled, False if not.
+        """
+        if "deny_platforms" in properties:
+            # setting can be Linux, Windows or Mac
+            curr_os = {"linux2": "Linux", "darwin": "Mac", "win32": "Windows"}[sys.platform]
+            if curr_os in properties["deny_platforms"]:
+                # not enabled on this platform
+                return False
+
+        return True
+
+    @classmethod
+    def is_compatible(cls, data):
+        """
+        Determines if the given data is compatible.
+
+        :param dict data: Serialized data
+        :returns: True if the given data can be loaded, False if not.
+        """
+        try:
+            return data.get("generation") == cls.FORMAT_GENERATION
+        except Exception as e:
+            return False
 
     @classmethod
     def create(cls, external_configuration, data, entity_id):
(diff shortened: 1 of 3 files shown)


python/external_config/external_command.py
```diff
@@ -59,13 +61,44 @@ def serialize_command(cls, entity_type, command_name, properties):
             "group_default": properties.get("group_default") or False,
 
             # special for shotgun
-            "deny_permissions": properties.get("deny_permissions"),
-            "deny_platforms": properties.get("deny_platforms"),
```
sgsteph: Just checking: `deny_platforms` is not serialized anymore and now fully dynamic, right?

python/external_config/external_command.py
```diff
@@ -59,13 +61,44 @@ def serialize_command(cls, entity_type, command_name, properties):
...
+        Checks toolkit command properties to determine
+        if a command is enabled on the current os or Not.
+
+        :param dict properties: Properties dictionary
+            as returned by the ``Engine.commands`` dictionary property.
+        :return: True if enabled, False if not.
+        """
+        if "deny_platforms" in properties:
+            # setting can be Linux, Windows or Mac
+            curr_os = {"linux2": "Linux", "darwin": "Mac", "win32": "Windows"}[sys.platform]
```
sgsteph: Don't we have a helper for this? I remember having similar code somewhere else, so could be the occasion to have a standard single place to convert `sys.platform` to the names we handle. And this could be cached when this file is imported, rather than being done here.

python/external_config/external_command.py
```diff
@@ -59,13 +61,44 @@ def serialize_command(cls, entity_type, command_name, properties):
...
+
+    @classmethod
+    def is_compatible(cls, data):
+        """
+        Determines if the given data is compatible.
+
+        :param dict data: Serialized data
+        :returns: True if the given data can be loaded, False if not.
+        """
+        try:
```
sgsteph: Is it some kind of "ask for forgiveness" pattern? If so, what is the exception we expect here? We use a `get` so we can't have a `KeyError`, so it seems to only error could come from `data` not being a dictionary?

python/external_config/external_command.py
```diff
@@ -125,6 +168,10 @@ def __init__(
         :param str entity_type: Associated entity type
         :param int entity_id: Associated entity id
         :param str pipeline_config_name: Associated pipeline configuration name
+        :param list sg_deny_permissions: (Shotgun specific, legacy) List of permission
+            groups to eclude this action from.
```
sgsteph: exclude?

python/external_config/external_command.py
```diff
@@ -221,13 +271,55 @@ def pipeline_configuration_name(self):
...
+    def group(self):
+        """
+        Group command belongs to or None if not defined.
+        """
+        return self._group
+
+    @property
+    def is_group_default(self):
+        """
+        True if this command is a default action for a group.
```
sgsteph: Not sure I know what a default action for a group is, could you add a tiny explanation?

python/external_config/external_command.py
```diff
@@ -221,13 +271,55 @@ def pipeline_configuration_name(self):
...
+
+        Returns a list of Shotgun permission groups (as strings)
+        where this command is not appropriate.
+        """
+        return self._sg_deny_permissions or []
+
+    @property
+    def support_shotgun_multiple_selection(self):
+        """
+        Legacy flag indicated by some older Toolkit apps,
```
sgsteph: Just checking: we say legacy, does that mean that we don't plan to support this anymore and this is kept for backward compatibility?

python/external_config/scripts/external_runner.py
```diff
@@ -304,6 +308,16 @@ def main():
     """
     task_runner = QtTaskRunner(main)
 
+    # For qt5, we may get this error:
+    #
+    # RuntimeError: Qt WebEngine seems to be initialized from a plugin.
+    # Please set Qt::AA_ShareOpenGLContexts using QCoreApplication::setAttribute
+    # before constructing QGuiApplication.
+    if hasattr(qt_importer.QtCore.Qt, "AA_ShareOpenGLContexts"):
```
sgsteph: Do we need a `QtGui` application here? Could this be solved by using a `QtCore` application?

python/external_config/external_command.py
```diff
@@ -59,13 +61,44 @@ def serialize_command(cls, entity_type, command_name, properties):
...
+            "sg_supports_multiple_selection": properties.get("supports_multiple_selection"),
         }
 
         return data
 
+    @classmethod
+    def enabled_on_current_os(cls, properties):
+        """
+        Checks toolkit command properties to determine
+        if a command is enabled on the current os or Not.
```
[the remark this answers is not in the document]
manneohrstrom: fixed!

python/external_config/external_command.py
```diff
@@ -59,13 +61,44 @@ def serialize_command(cls, entity_type, command_name, properties):
...
+
+        :param dict properties: Properties dictionary
+            as returned by the ``Engine.commands`` dictionary property.
+        :return: True if enabled, False if not.
```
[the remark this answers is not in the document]
manneohrstrom: fixed

python/external_config/external_command.py
manneohrstrom: good point! the caches are assumed to be per-os, so we are 'baking' the state of the current os into the cache. adding a comment to clarify!

python/external_config/external_command.py
manneohrstrom: fixed!

python/external_config/external_command.py
manneohrstrom: sorry, absolutely. will add.

python/external_config/external_command.py
manneohrstrom: `ShotgunPath` doesn't actually provide any good services for this, i looked into it - and i am not sure it's worth adding.

python/external_config/external_command.py
manneohrstrom: Changing to `KeyError`!

python/external_config/external_command.py
manneohrstrom: good point. at some point we considered it legacy, but in recent docs for `register_command` it's not legacy so i'll remove that here.

python/external_config/scripts/external_runner.py
manneohrstrom: We always launch a `QTGui` application right now - perhaps in the future we can restrict this, but for the time being that seems to be the simplest approach for launching

PR #88, opened 2018-03-27
#47089 Adjustments to signal policy

manneohrstrom:
Tweak to the behaviour of the `request_configurations()` method -- if called multiple times in rapid succession, only a single signal will be emitted. Any queued up or in-progress signals are basically discarded when `request_configurations()` is called.


manneohrstrom:
@sgsteph approved this via slack. I'm merging :)

PR #90, opened 2018-05-20
SG-6700 Fixes bug for bg task manager assignment.

manneohrstrom:
See code. This fixes an issue where the bg task manager was actually never properly passed to the shotgunmodel base classes from the hierarchy widget, if it was provided. This resulted in a separate bg task manager being allocated, learning to resource leaks and unclean shutdown.

Big shoutout and thanks to @akitisa-epic for reporting and submitting these!

### DIFF
--- a/python/shotgun_model/shotgun_hierarchy_model.py
+++ b/python/shotgun_model/shotgun_hierarchy_model.py
@@ -94,7 +94,11 @@ class ShotgunHierarchyModel(ShotgunQueryModel):
             supplied arg would look like: ``include_root="Project Publishes"``.
             If ``include_root`` is `None`, no root item will be added.
         """
-        super(ShotgunHierarchyModel, self).__init__(parent, bg_task_manager)
+        super(ShotgunHierarchyModel, self).__init__(
+            parent,
+            bg_load_thumbs=True,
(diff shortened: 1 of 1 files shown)


python/shotgun_model/shotgun_hierarchy_model.py
```diff
@@ -94,7 +94,11 @@ def __init__(self, parent, schema_generation=0, bg_task_manager=None, include_ro
             supplied arg would look like: ``include_root="Project Publishes"``.
             If ``include_root`` is `None`, no root item will be added.
         """
-        super(ShotgunHierarchyModel, self).__init__(parent, bg_task_manager)
+        super(ShotgunHierarchyModel, self).__init__(
+            parent,
+            bg_load_thumbs=True,
```
josh-t: Does this change the behavior or is this the default? Any concerns here?

manneohrstrom: possibly but this is a good thing. Previously it would pass the bg task manager class instance into a bool and that would be the state of bg_load_thumbnails - so in the cases where a task manager was explicitly passed in, there would be thumbs otherwise not.

PR #91, opened 2018-06-08
SG-6836: Fixes a bug that manifested when a model hard_refresh was called

thebeeland:
During the hard refresh of a ShotgunQueryModel, the data handler's cache would be cleared out, and then the refresh process would request an update of the model's data. That call would fail, as it raised when no cache existed for the data handler. Instead, we now treat the missing cache kindly and just continue on with the update, which itself populates a fresh cache.

### DIFF
--- a/python/shotgun_model/shotgun_query_model.py
+++ b/python/shotgun_model/shotgun_query_model.py
@@ -268,10 +268,18 @@ class ShotgunQueryModel(QtGui.QStandardItemModel):
         # Clear ourselves, preserving the data handler so that we can then
         # refresh. Clearing the model here ensures we don't end up with
         # duplicate items once the cache is cleared and repopulated.
-        self.clear(remove_data_handler=False)
-
-        # request a reload
-        self._refresh_data()
+        #
+        # Block all signals before we clear the model otherwise downstream
+        # proxy objects could cause crashes.
+        signals_blocked = self.blockSignals(True)
+        try:
+            # Clear all internal memory storage.
+            self.clear(remove_data_handler=False)
+            self._refresh_data()
+        finally:
+            # Reset the state of signal blocking.
+            self.blockSignals(signals_blocked)
+            self.modelReset.emit()
 
     def is_data_cached(self):
         """
--- a/python/shotgun_model/shotgun_query_model.py
+++ b/python/shotgun_model/shotgun_query_model.py
@@ -170,10 +170,14 @@ class ShotgunQueryModel(QtGui.QStandardItemModel):
     ############################################################################
     # public methods
 
-    def clear(self):
+    def clear(self, remove_data_handler=True):
         """
         Removes all items (including header items) from the model and
         sets the number of rows and columns to zero.
+
+        :param bool remove_data_handler: Whether to remove the reference to the
+            current data handler object. If False, the data handler will be
+            retained, which will allow the model to be repopulated.
         """
         # clear thumbnail download lookup so we don't process any more results:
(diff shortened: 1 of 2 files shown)


python/shotgun_model/shotgun_query_model.py
```diff
@@ -170,10 +170,14 @@ def __init__(self, parent, bg_load_thumbs, bg_task_manager=None):
     ############################################################################
     # public methods
 
-    def clear(self):
+    def clear(self, remove_data_handler=True):
         """
         Removes all items (including header items) from the model and
         sets the number of rows and columns to zero.
+
+        :param bool remove_data_handler: Whether to remove the reference to the
```
manneohrstrom: Curious why this is needed? Is it an optimization? It feels like we are exposing an internal implementation detail in the public interface - `data_handler` is an internal thing which I *think* is nothing we want to tell our users about (and may want to change in the future).

If it is an optimization, i think we should try to name it differently - we should also consider whether it's needed at all? The `hard_refresh` operation is a very low frequency operation.

python/shotgun_model/shotgun_query_model.py
```diff
@@ -260,8 +264,22 @@ def hard_refresh(self):
...
+        # Clear ourselves, preserving the data handler so that we can then
+        # refresh. Clearing the model here ensures we don't end up with
+        # duplicate items once the cache is cleared and repopulated.
+        #
+        # Block all signals before we clear the model otherwise downstream
+        # proxy objects could cause crashes.
+        signals_blocked = self.blockSignals(True)
+        try:
+            # Clear all internal memory storage.
+            self.clear(remove_data_handler=False)
```
manneohrstrom: i think of the `hard_refresh` almost as a helper method -- we want to clear caches on disk, and reset the model so that next time we load data, we start fresh. It seems to me that now we have a 'special' contract, meaning that in order for this to work we have to pass a special `remove_data_handler` flag to the clear method to 'clear the model in a special way'. It's not clear to me why! Do we need a special way to clear it? Would just `self.clear()` work?

python/shotgun_model/shotgun_query_model.py
```diff
@@ -170,10 +170,14 @@ def __init__(self, parent, bg_load_thumbs, bg_task_manager=None):
     ############################################################################
     # public methods
 
-    def clear(self):
+    def clear(self, remove_data_handler=True):
```
manneohrstrom: This method shadows the built-in http://doc.qt.io/archives/qt-4.8/qstandarditemmodel.html#clear - i am not sure it makes sense to extend these qt model interface methods with our own parameters.

thebeeland: I've refactored the fix here to no longer change the class's public interface. You're absolutely right the we should try to avoid that wherever possible!

python/shotgun_model/shotgun_query_model.py
thebeeland: This has been refactored to take the small amount of work needed from clear() and use it directly instead of trying to bend clear() to my will. ;)

Hopefully it'll look more like what you're hoping for now!

python/shotgun_model/shotgun_query_model.py
thebeeland: This has been refactored to remove the need for the new argument.

thebeeland:
Refactored in #92

manneohrstrom:
Thanks @thebeeland for being patient with me! :)

PR #93, opened 2018-08-22
SG-7486: Handles inaccessible configs on disk gracefully.

thebeeland:
Previous to this change, if all PC entities pointed to configs on disk that weren't accessible from the host, no response would be sent to the web app requesting menu actions via WSS. Instead, we now thoroughly warn and use the fallback configuration.

### DIFF
--- a/python/external_config/external_config_loader.py
+++ b/python/external_config/external_config_loader.py
@@ -12,7 +12,7 @@ import sgtk
 from sgtk.platform.qt import QtCore, QtGui
 from .configuration_state import ConfigurationState
 from . import file_cache
-from .errors import ExternalConfigNotAccessibleError, ExternalConfigParseError
+from .errors import ExternalConfigParseError
 from . import config
 
 logger = sgtk.platform.get_logger(__name__)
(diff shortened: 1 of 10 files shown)


thebeeland:
@manneohrstrom I've implemented the logic necessary to raise an error message to the web app in this situation. The code is in this PR, but there's associated logic that was added in tk-desktop2 [here](https://github.com/shotgunsoftware/tk-desktop2/pull/11). Apologies for reusing an existing PR, but I'm trying to roll a ton of tickets together into a single branch so it's easy for Francis to QA everything in one go, and that's the branch I'm using to integrate everything into. I've put a couple of messages at the points in the diff where you need to look for this. It's a relatively small change.

python/external_config/external_config_loader.py
```diff
@@ -39,7 +41,7 @@ class ExternalConfigurationLoader(QtCore.QObject):
 
     # signal emitted to indicate that an update has been detected
     # to the pipeline configurations for a project
-    configurations_loaded = QtCore.Signal(int, list)  # project_id, list of configs
+    configurations_loaded = QtCore.Signal(int, list, tuple)  # project_id, list of configs, error
```
manneohrstrom: would it make sense to have a separate signal to indicate failure?

Also thinking if we rather than passing a string from this library should pass an error state/enum or something and then let desktop turn that into a response for the UI (in our case a string). Thinking about localization but also thinking this might give the caller a lot more flexibility in how errors are communicated to the user. With the string, all that the client code knows is that something went wrong. We already have a  `ExternalConfigNotAccessibleError` exception type bubbling up, so we got a good idea of the type of error.

PR #96, opened 2018-09-04
SG-4847: Supports executing commands with multiple entity ids as their input.

thebeeland:
Allows a list of entity ids to be passed in when executing a command. This list is then passed on to any call of `engine.execute_old_style_command` to allow for support of legacy commands that can run on multiple entities at the same time.

### DIFF
--- a/python/external_config/external_command.py
+++ b/python/external_config/external_command.py
@@ -297,7 +297,76 @@ class ExternalCommand(object):
         """
         return self._tooltip
 
-    def execute(self, pre_cache=False, entity_ids=None):
+    def execute(self, pre_cache=False):
+        """
+        Executes the external command in a separate process.
+
+        .. note:: The process will be launched in an synchronous way.
+            It is recommended that this command is executed in a worker thread::
+
+                # execute external command in a thread to not block
+                # main thread execution
+                worker = threading.Thread(target=action.execute)
+                # if the python environment shuts down, no need
+                # to wait for this thread
+                worker.daemon = True
+                # launch external process
+                worker.start()
+
+        :param bool pre_cache: If set to True, starting up the command
+            will also include a full caching of all necessary
+            dependencies for all contexts and engines. If set to False,
+            caching will only be carried as needed in order to run
+            the given command. This is an advanced setting that can be
+            useful to set to true when launching older engines which don't
+            launch via a bootstrap process. In that case, the engine simply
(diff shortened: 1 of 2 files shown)


python/external_config/external_command.py
```diff
@@ -351,6 +356,7 @@ def execute(self, pre_cache=False):
                 engine_name=self._engine_name,
                 entity_type=self._entity_type,
                 entity_id=self._entity_id,
+                entity_ids=entity_ids or [self._entity_id], # Legacy support for multi-entity commands
```
manneohrstrom: do we need to pass both the `entity_id` and `entity_ids` parameters?

python/external_config/external_command.py
```diff
@@ -297,7 +297,7 @@ def tooltip(self):
         """
         return self._tooltip
 
-    def execute(self, pre_cache=False):
+    def execute(self, pre_cache=False, entity_ids=None):
```
manneohrstrom: I am not 100% sure here, but perhaps we should just have a separate method called `execute_multiple_selection()` or `execute_shotgun_multiple_selection()` (is this only possible via the shotgun engine?) - to separate out this special case more from the normal use case. 

I don't think it's a legacy case - i  think we call it that in other places in the code too - at some point it was meant to become the legacy case but it turned out having multiple selection was an important aspect of some shotgun actions yet at the same time not applicable in other cases. 

The fact that this is so different and separate suggests that it might be worth exploring an interface with separate methods. But maybe that complicates the calling side of things?

python/external_config/external_command.py
```diff
@@ -326,6 +326,80 @@ def execute(self, pre_cache=False):
         :raises: :class:`RuntimeError` on execution failure.
         :returns: Output from execution session.
         """
+        return self._execute(pre_cache)
+
+    def execute_on_multiple_entities(self, pre_cache=False, entity_ids=None):
+        """
+        Executes the external command in a separate process. This method
```
manneohrstrom: I like this approach a lot!! Thanks for trying it out!

In order to avoid the redundancy, perhaps we could just shorten the docstrings for `execute_on_multiple_entities` and `_execute` and have them refer back to `execute` for the details?

thebeeland:
This has been merged into #93 and will make its way from there.

python/external_config/external_command.py
thebeeland: 👍

PR #97, opened 2018-09-12
SG-7486: Reimplements how inaccessible configurations are communicated.

thebeeland:
This is a reimplementation of how we track inaccessible configurations found when getting configs, and how those are reported to the caller. Instead of passing error messages to the caller, we now track invalid configurations as an "invalid" external configuration object with an enum status that can be used to know WHY the configuration is invalid.

Currently, we only have enums for CONFIGURATION_READY and CONFIGURATION_INACCESSIBLE, but this can be expanded in the future if required.

### DIFF
--- a/python/external_config/config/config_base.py
+++ b/python/external_config/config/config_base.py
@@ -37,6 +37,10 @@ class ExternalConfiguration(QtCore.QObject):
     # grouping used by the background task manager
     TASK_GROUP = "tk-framework-shotgunutils.external_config.ExternalConfiguration"
 
+    # Status enums:
+    CONFIGURATION_READY = 1
+    CONFIGURATION_INACCESSIBLE = 2
+
     commands_loaded = QtCore.Signal(int, str, int, str, object, list)
     # Signal parameters:
     # 1. project_id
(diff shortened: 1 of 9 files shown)


python/external_config/config/config_base.py
```diff
@@ -201,7 +226,21 @@ def request_commands(self, project_id, entity_type, entity_id, link_entity_type,
...
             specified. This can be useful in backwards compatibility scenarios.
-        """
+
+        :raises: RuntimeError if this configuration's status does not allow for
+            commands requests.
+        """
+        # Make sure that we're a valid configuration. If we aren't, then we
+        # won't be able to successfully request commands. That being the case,
+        # we can raise here.
+        if not self.is_valid:
```
thebeeland: I'm having second thoughts about this, @manneohrstrom. I'm wondering if status should just be a property on the InvalidExternalConfiguration implementation, which itself raises if commands are requested from it. Then we can remove the status from the base external config entirely. Thoughts?

thebeeland: We'd then also no longer need the CONFIGURATION_READY status. In fact, we might not need the status enum at all since we only have one case right now. We could always add status back in the future if we end up with multiple possibilities for why a config is invalid. That would cut out some arguably-premature optimization here.

thebeeland:
@manneohrstrom I'm now questioning whether we need a status enum at all. We only have one possible cause for an invalid config right now, so maybe that's premature optimization? We can just rely on `is_valid` for now, and if we end up with more cases in the future we could introduce a status property and the necessary enums. Thoughts?

manneohrstrom:
@thebeeland sorry didn't see the commit before i did the CR. I imagines we had two invalid cases - one where the config cannot be found and one where it's not valid? Whatever you think is right, you have a much better sense of the details at this point.

PR #103, opened 2019-02-07
This is a workaround for https://jira.autodesk.com/browse/SHOT-2806

000paradox000:
Toolkit expects the following code

from PySide2 import QtGui
i = QtGui.QStandardItem()
i.setData(1, 1)
print i.data(1)._class_

to print out <type 'int'>, but in the case of VRed it gives back <type 'long'>.

### DIFF
--- a/python/shotgun_model/util.py
+++ b/python/shotgun_model/util.py
@@ -149,10 +149,7 @@ def sanitize_qt(val):
         return new_val
 
     elif isinstance(val, long):
-        try:
-            val = int(val)
-        except ValueError:
-            pass
+        val = int(val)
         return val
     else:
         return val        
--- a/python/shotgun_model/util.py
+++ b/python/shotgun_model/util.py
@@ -148,15 +148,10 @@ def sanitize_qt(val):
             new_val[safe_key] = safe_val
         return new_val
 
-    """
-    This is a workaround for the JIRA ticket 
-    https://jira.autodesk.com/browse/SHOT-2806
-    QT Version: 5.9.5
-    PySide Version: 5.9.0a1
-    VRED version: 2020
-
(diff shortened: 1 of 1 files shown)


python/shotgun_model/util.py
```diff
@@ -148,6 +148,12 @@ def sanitize_qt(val):
             new_val[safe_key] = safe_val
         return new_val
 
+    elif isinstance(val, long):
+        try:
+            val = int(val)
```
jfboismenu: I just tried, and `int(1000000000000000000000000000000000000000000000000)` will return a long, so no need to try/except

000paradox000: Thanks I just removed try - except

python/shotgun_model/util.py
```diff
@@ -148,6 +148,9 @@ def sanitize_qt(val):
             new_val[safe_key] = safe_val
         return new_val
 
+    elif isinstance(val, long):
```
manneohrstrom: we need to add some comments here to explain why we have to do this! If possible please include info about versions of pyside, QT etc. This will help maintain the code in the future and help people debug in different DCCs when issues arise! Thanks!! :)

000paradox000: Hi @manneohrstrom I added some comments

manneohrstrom: thanks great! Sorry for being nitpicky, could you tweak it a bit:
- Remember this is a public repo, everyone can see, so our jira ticket isn't really helpful for people (they cannot access our bug tracker!). I suggest just removing that. People curious about it can always follow the PR trail and find the ticket id that way.
- Please change it into a `# comment` rather than a string - that's how we normally do things!

Thanks!!!!

PR #104, opened 2019-02-21
VMR-1680: Add icon property to ExternalCommand class

nmontmarquette:
The Shotgun Create application needed a quick and easy way to access a command's icon.


nmontmarquette:
I'm using the serialized ExternalCommand object from within the ` view-master/src/ui/QtoolkitDriver.cpp` to package an extra icon field to the UI (React) code.

python/external_config/external_command.py
```diff
@@ -419,6 +432,7 @@ def _execute(self, pre_cache=False, entity_ids=None):
                 supports_multiple_selection=self._sg_supports_multiple_selection,
                 pre_cache=pre_cache,
                 pythonpath=current_pypath,
+                # Todo: Do I need to add icon here as well?
```
nmontmarquette: Also, I wasn't sure about that one

manneohrstrom: no need to pass the icon here!

nmontmarquette: Good I'll clean that one line up then

nmontmarquette:
Do you require me to add test? I believe this file has none.

python/external_config/external_command.py
```diff
@@ -237,6 +243,13 @@ def display_name(self):
         """
         return self._display_name
 
+    @property
+    def icon(self):
+        """
+        The path to a square png icon file representing this item
```
manneohrstrom: we should add : "or None if no icon is associated"

nmontmarquette: Should the property be called icon_256 ? for consistency with overall TK interface ?

nmontmarquette: Not sure whether it's None, I really just copy over what I get from the `data` dict, I would assume it might be empty string

thebeeland: I'd prefer to stick with "icon" instead of "icon_256"

thebeeland: My guess would be a `None` and not an empty string, but I'm not 100% sure on that.

tests/external_config/__init__.py
```diff
@@ -7,3 +7,91 @@
...
 # agreement to the Shotgun Pipeline Toolkit Source Code License. All rights
 # not expressly granted therein are reserved by Shotgun Software Inc.
+
+import sys
+import os
+
+from mock import Mock
+
+# import the test base class
+test_python_path = os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "python"))
```
jfboismenu: No need to add to sys.path for. The toolkit unit test framework does it for you.

tests/external_config/test_external_command.py
```diff
@@ -0,0 +1,131 @@
...
+    @property
+    def ec_data(self):
+        return self._data
+
+    @property
+    def ec(self):
+        return self._external_command
+
+    def test_properties(self):
+        self.assertEquals(self.ec.pipeline_configuration_name, self._pipeline_configuration_id)
```
jfboismenu: Missing docstring.

tests/external_config/test_external_command.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertEquals(self.ec.is_group_default, self.ec_data["group_default"])
+        if self.ec_data["sg_deny_permissions"]:
+            self.assertEquals(self.ec.excluded_permission_groups_hint, self.ec_data["sg_deny_permissions"])
+        else:
+            self.assertEquals(self.ec.excluded_permission_groups_hint, [])
+        self.assertEquals(self.ec.support_shotgun_multiple_selection, self.ec_data["sg_supports_multiple_selection"])
+        self.assertEquals(self.ec.tooltip, self.ec_data["tooltip"])
+        self.assertTrue(len(repr(self.ec)) > 0)
+
+    def test_serialize_deserialize(self):
```
jfboismenu: Missing docstring.

tests/external_config/test_external_command.py
```diff
@@ -0,0 +1,131 @@
...
+        # Creating the external command object
+        self._external_command = self.external_config.ExternalCommand.create(
+            self.external_config_loader,
+            self.ec_data,
+            "Task"
+        )
+
+    @property
+    def ec_data(self):
+        return self._data
```
jfboismenu: Missing docstring.

tests/external_config/test_external_command.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @property
+    def ec(self):
```
jfboismenu: Missing docstring.

PR #108, opened 2019-10-23
SG-13267: Python 3 port

jfboismenu:
Ports the framework to Python 3. Most of the fixes were decisions about sticking with the iterable methods. Lots of tests needed to be fixed so that they were more tolerant of variable ordering. A lot of this code uses dicts and sets whose ordering is not only different than Python 3, but can actually be different between two invocations of the Python 3 interpreter!

I've kept the travis build around, as it provides for coverage for Python 2.7, which we do not have at the moment in Azure due to Azure providing a Python not compatible with the PySide versions we use.


.travis.yml
```diff
@@ -6,18 +6,12 @@ python:
...
-    - pushd /tmp/tk-core
-    - git checkout master
-    - popd
+    - git clone --branch SG-13264-python-3-compat --depth 1 https://github.com/shotgunsoftware/tk-core /tmp/tk-core
     # install pyside - from https://stackoverflow.com/questions/24489588
     - sudo apt-get install libqt4-dev
     - pip install PySide --no-index --find-links https://parkin.github.io/python-wheelhouse/;
     # Travis CI servers use virtualenvs, so we need to finish the install by the following
     - python ~/virtualenv/python${TRAVIS_PYTHON_VERSION}/bin/pyside_postinstall.py -install
-    - pip install coverage
```
jfboismenu: Azure deals with this now.

.travis.yml
```diff
@@ -6,18 +6,12 @@ python:
...
-    - pip install coveralls
 
 # command to run tests
-script: /tmp/tk-core/tests/run_tests.sh --test-root tests --with-coverage
-after_success: coveralls
+script: python /tmp/tk-core/tests/run_tests.py --test-root tests
```
jfboismenu: The python 3 on tk-core tries to run tests with both Python 2 and 3, but we only want to run with 2.

framework.py
```diff
@@ -122,21 +124,15 @@ def _remove_old_cached_data(self, grace_period, *cache_locations):
...
-                                     modification should be kept around.
+                                 modification should be kept around.
         :param cache_locations: A list of cache locations to clean up.
         :raises: ValueError if the grace period is smaller than one day.
         """
         if grace_period < 1:
             raise ValueError(
                 "Invalid grace period value %d, it must be a least 1" % grace_period
             )
-        delta = datetime.timedelta(days=grace_period)
```
jfboismenu: This unintuitive piece of code caused an error with Python 3.7 due to the use of a true division. I've simplified the code since it introduced errors in Python 3 anyway. Both this code and it's test now use the much simpler assumption that `24 * 3600` is a full day. We're trying to roughly count days, not be super precise to the millisecond about it anyway.

python/shotgun_model/data_handler_cache.py
```diff
@@ -130,7 +131,7 @@ def get_children(self, parent_uid):
             cache_node = self._cache[self.CACHE_BY_UID].get(parent_uid)
 
         if cache_node:
-            for item in cache_node[self.CACHE_CHILDREN].itervalues():
+            for item in six.itervalues(cache_node[self.CACHE_CHILDREN]):
```
jfboismenu: Caches can be really, really big, so we're sticking to itervalues instead of just calling `.values()`

tests/external_config/test_external_command.py
```diff
@@ -78,33 +78,34 @@ def test_properties(self):
...
             self.ec.pipeline_configuration_name, self._pipeline_configuration_id
         )
         # Yes, the prop. name and dict key are different
-        self.assertEquals(self.ec.system_name, self.ec_data["callback_name"])
-        self.assertEquals(self.ec.engine_name, self.ec_data["engine_name"])
-        self.assertEquals(self.ec.display_name, self.ec_data["display_name"])
-        self.assertEquals(self.ec.group, self.ec_data["group"])
-        self.assertEquals(self.ec.icon, self.ec_data["icon"])
-        self.assertEquals(self.ec.is_group_default, self.ec_data["group_default"])
+        self.assertEqual(self.ec.system_name, self.ec_data["callback_name"])
```
jfboismenu: These generated warnings in pytest because they are deprecated, so I fixed the calls.

jfboismenu: You can skip this file, it's all about fixing that warning.

tests/shotgun_model/test_data_handler_find.py
```diff
@@ -171,26 +181,29 @@ def test_updates(self):
         # we are getting a diff of two added nodes back.
         self.assertEqual(len(diff), 4)
 
+        # Items will be sorted as /Character, /Prop, 1234 and 3333.
```
jfboismenu: Do a deterministic sort of the entries so Python 3's output is predictable.

framework.py
willis102: But what if the number of days in the year or hours in the day changes??  😆 (This is way easier to read, makes sense.)

python/shotgun_globals/cached_schema.py
```diff
@@ -255,14 +255,13 @@ def _check_schema_refresh(self, entity_type=None, field_name=None, project_id=No
         # or not. Later on, when we have the ability to check the current
         # metaschema generation via the shotgun API, this can be handled in a
         # more graceful fashion.
-        if (
-            not self._is_schema_loaded(project_id)
-            and project_id not in self._sg_schema_query_ids.values()
+        if not self._is_schema_loaded(project_id) and project_id not in list(
```
willis102: I could be wrong, but I don't believe the list instantiation here is necessary?

python/shotgun_globals/cached_schema.py
```diff
@@ -295,13 +294,12 @@ def _check_status_refresh(self, project_id=None):
         current_project_id = self._get_current_project_id()
         project_id = project_id or current_project_id
 
-        if (
-            not self._is_status_loaded(project_id)
-            and project_id not in self._sg_status_query_ids.values()
+        if not self._is_status_loaded(project_id) and project_id not in list(
```
willis102: Same as above

tests/shotgun_model/test_data_handler_find.py
willis102: 👍

python/settings/user_settings.py
```diff
@@ -134,12 +167,13 @@ def retrieve(self, name, default=None, scope=SCOPE_GLOBAL):
         self.__fw.log_debug("User Settings Manager: Retrieving %s" % full_name)
 
         try:
-            raw_value = sanitize_qt(self.__settings.value(full_name))
```
jfboismenu: Settings were already sanitized on their way in, so no need to sanitize on their way out.

python/external_config/scripts/external_runner.py
```diff
@@ -14,9 +14,9 @@
 import imp
 import sys
 import errno
-import cPickle
 import inspect
 import traceback
+from tank_vendor.shotgun_api3.lib import six
```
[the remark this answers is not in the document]
willis102: I think hound is right about this, no?

python/settings/user_settings.py
```diff
@@ -9,8 +9,9 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import sgtk
-import urlparse
-import cPickle as pickle
+from tank_vendor.shotgun_api3.lib import six
+import json
```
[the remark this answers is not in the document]
willis102: And here?

PR #111, opened 2020-01-28
SG-13276: Fixed task priority sorting

jfboismenu:
The workfiles 2 app passes in a priotity for it's threads, unlike most of the rest of our code, so we didn't catch this until now.

In Python 2, comparing `int` and `NoneType` was possible, but it is not permitted in Python 3, so we're making sure that priorities are always numerical now.

### DIFF
--- a/python/task_manager/worker_thread.py
+++ b/python/task_manager/worker_thread.py
@@ -52,8 +52,8 @@ class WorkerThread(QtCore.QThread):
         """
         Shut down the thread and wait for it to exit before returning
         """
-        self._results_dispatcher = None
         with self._mutex:
+            self._results_dispatcher = None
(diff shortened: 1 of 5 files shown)


python/task_manager/worker_thread.py
```diff
@@ -52,8 +52,8 @@ def shut_down(self):
         """
         Shut down the thread and wait for it to exit before returning
         """
-        self._results_dispatcher = None
         with self._mutex:
+            self._results_dispatcher = None
```
jfboismenu: This actually fixed a small race condition when tests finalized. Sometimes the dispatcher would become None in other parts of the code while the mutex was taken, which is a no-no, considering it's role is to provide thread-safe access to it!

PR #113, opened 2020-04-24
Improve Python3 support

geffrak:
This pull request avoid this issue:

```
".../python/external_config/external_config_loader.py", line 169, in request_configurations
    for task_id in self._task_ids.keys():
RuntimeError: dictionary changed size during iteration
```

The fix is pretty straight forward.

### DIFF
--- a/python/external_config/external_config_loader.py
+++ b/python/external_config/external_config_loader.py
@@ -166,7 +166,7 @@ class ExternalConfigurationLoader(QtCore.QObject):
         # in rapid succession.
         #
         # note: not using a generator since we are deleting in the loop
-        for task_id in self._task_ids.keys():
+        for task_id in list(self._task_ids.keys()):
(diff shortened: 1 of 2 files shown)


python/external_config/external_config_loader.py
```diff
@@ -166,7 +166,7 @@ def request_configurations(self, project_id):
         # in rapid succession.
         #
         # note: not using a generator since we are deleting in the loop
-        for task_id in self._task_ids.keys():
+        for task_id in list(self._task_ids.keys()):
```
jfboismenu, line 169: Any chance we can add a test for this function? We clearly didn't have coverage and this is a vital part of browser integration.

jfboismenu: You'll likely mock things a bit so it doesn't spawn a subprocess.

PR #114, opened 2020-05-01
Unset PYTHONHOME in external_runner.py

geffrak:
Some versions of Python sets the PYTHONHOME environment variable and it might conflict with some DCC since it affects the way Python is initialized. 

Ex. Maya and Nuke fail to start if the variable is set.

This change is required because upgrading the python interpreters in Shotgun Create introduces the behaviour of setting that variable.


geffrak:
This is not required anymore since I found a workaround that I can apply in Shotgun Create. While the issue is still valid, the scope of the changes are more controlled if I do it in the Create codebase.

PR #118, opened 2020-06-24
SG-17970 user settings python 3 pickle issue fix

pscadding:
We were seeing issues in Shotgun Desktop Python 3 with storing a datetime object as a setting.

```
2020-06-12 10:53:40,147 [ WARNING] Error storing user setting 'asite.shotgunstudio.com:project_recent_apps.215'. Error details: 'utf-8' codec can't decode byte 0xe4 in position 113: invalid continuation byte 
```

*Brief background: The settings utility tries to pickle objects into strings before passing them to QSettings to store, so that we have control over how that happens.* 

The problem was that it was using the `tk-core` pickle implementation which tries to ensure that the pickled object is a `utf-8` string. However for some reason, that I don't fully understand, it appears to need to be a `latin1` str. So rather than changing core we opted to handle the pickling inside the settings code.

### DIFF
--- a/python/settings/user_settings.py
+++ b/python/settings/user_settings.py
@@ -9,10 +9,12 @@
 # not expressly granted therein are reserved by Shotgun Software Inc.
 
 import sgtk
-from tank_vendor import six
 from sgtk.platform.qt import QtCore
 from sgtk import TankError
 from ..shotgun_model import sanitize_qt
+
+from tank_vendor import six
+from tank_vendor.six.moves import cPickle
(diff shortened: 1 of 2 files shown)


pscadding:
I've renamed the branch:
https://github.com/shotgunsoftware/tk-framework-shotgunutils/tree/fix-for-user-settings-pickle-issues

I'm going to adjust tk-desktop to stop it storing a datetime object, and Qpoint object, and park this.

PR #119, opened 2020-07-21
SG-18486 Now suppresses warnings that were popping up in browser integration.

pscadding:
This suppresses the libpng and imp deprecation warnings that are shown in the browser integration when using SG Create.

The imp warning is only logged when the script that is directly run in python has an imp import in it. Importing other modules from that script that contain imp import won't trigger the warning. However in this case the `external_runner.py` script is executed in a new Python instance, and contains an imp import in it, so we must suppress the warning.

### DIFF
--- a/python/external_config/scripts/external_runner.py
+++ b/python/external_config/scripts/external_runner.py
@@ -156,7 +156,7 @@ def _handle_qt_warnings():
     """
     This will suppress the libpng warnings, but allow any
     other warnings or errors from QT to print.
-    We do this because warning by default get printed to stderr
+    We do this because by default warnings get printed to stderr
     and so they get displayed in the browser when running actions.
     """
 
--- a/python/external_config/scripts/external_runner.py
+++ b/python/external_config/scripts/external_runner.py
@@ -11,11 +11,20 @@
 from __future__ import print_function
 import os
 import re
-import imp
 import sys
 import errno
 import inspect
 import traceback
+
+# Until we remove the use of imp from this code,
+# we must suppress the warning here as it will pop up in the Shotgun browser
+# integration as a message box, when running in Python 3.4 >.
+import warnings
+
+with warnings.catch_warnings():
+    warnings.filterwarnings("ignore", category=DeprecationWarning)
+    import imp
+
 
 # handle imports
 path_to_sgtk = sys.argv[1]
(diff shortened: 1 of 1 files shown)


python/external_config/scripts/external_runner.py
```diff
@@ -143,6 +152,44 @@ def execute_command(self):
...
+            # This is probably a debug or info message so just print these normally.
+            print(msg_string)
+
+    # Add a message handler so we can suppress the warnings about libpng.
+    try:
+        # QT 4
+        qt_importer.QtCore.qInstallMsgHandler(handler)
+    except AttributeError:
+        # QT 5
+        qt_importer.QtCore.qInstallMessageHandler(handler)
```
pscadding, line 190: Maybe this stuff should be moved into core as a utility method? could be useful?

python/external_config/scripts/external_runner.py
```diff
@@ -143,6 +152,44 @@ def execute_command(self):
...
+        if msg_string == "libpng warning: iCCP: known incorrect sRGB profile":
+            return
+
+        if msg_type in [
+            qt_importer.QtCore.QtMsgType.QtWarningMsg,
+            qt_importer.QtCore.QtMsgType.QtCriticalMsg,
+            qt_importer.QtCore.QtMsgType.QtFatalMsg,
+        ]:
+            # By default Qt would usually print these to stderr so we should do the same.
+            print(msg_string, file=sys.stderr)
```
geffrak, line 179: Not that I care that much, but we lose the context in Qt 5 with that way of printing it.

geffrak: This might be fixed by something like printing the result of args[1:] joined in a nice way.

pscadding: Hmm good point. Although if I comment out the libpng suppressing bit, it still prints that warning just the same as if I didn't take over the message handler (with QT 5):
`libpng warning: iCCP: known incorrect sRGB profile`

So for that warning at least it seems the default behaviour doesn't make use of the context.
Looking at the [example in the QT docs](https://doc.qt.io/qt-5/qtglobal.html#qInstallMessageHandler) they use the context to mention the line number. Perhaps this just doesn't make sense in a Python context where the line number would presumably be in the compiled C++ code?

geffrak: Good enough then!

python/external_config/scripts/external_runner.py
```diff
@@ -143,6 +152,44 @@ def execute_command(self):
...
+    """
+
+    def handler(*args):
+        # We handle the args this way since Qt 5 passes 3 args, type, context and message,
+        # Where as Qt 4 only passes 2, type, and message.
+        msg_type = args[0]
+        msg_string = args[-1]
+
+        # Suppress this warning.
+        if msg_string == "libpng warning: iCCP: known incorrect sRGB profile":
```
jfboismenu, line 170: MIND = BLOWN.

I was sure this was coming from Qt itself. So cool.

python/external_config/scripts/external_runner.py
jfboismenu: Let's wait and see if this becomes a pattern first.

PR #121, opened 2020-10-26
SG-19850 - Make hashes generation deterministic

geffrak:
Hashes generation relied on the `hash` method, which is not deterministic in Python 3.


geffrak:
I will add tests tomorrow.

python/external_config/configuration_state.py
```diff
@@ -194,7 +198,12 @@ def get_hash(self):
             # note: there *may* be a bug when deleting items out of a shotgun
             #       model in certain cases, so in order to ensure we
             #       get a correct representation, include entity_ids in the hash.
-            return hash(str(sg_data + self.entity_ids))
```
jfboismenu, line 197: Ah! I wasn't aware that hash wasn't deterministic! Seems like we've read both the same comment on Stack Overflow!
https://stackoverflow.com/questions/27954892/deterministic-hashing-in-python-3

However, your output it STILL not going to be deterministic here. That's because we generate a string out of the `sg_data`  dictionary, which is NOT deterministic. This is why [in other places we create a JSON document with sorted keys](https://github.com/shotgunsoftware/tk-core/blob/master/python/tank/context.py#L244) first to ensure a deterministic output.

For sake of conformity, I'd recommend using hashlib's MD5 like we do [in multiple places](https://github.com/shotgunsoftware/tk-framework-shotgunutils/blob/master/python/shotgun_model/shotgun_hierarchy_model.py#L602).

jfboismenu (changes requested):
TIL `hash` wasn't deterministic. 🤯 

Unfortunately, your code still has some issues. See comments.

python/external_config/configuration_state.py
jfboismenu: About the ordering of keys:
Python 2.7 has no guarantee that the ordering of keys is going to be deterministic. Starting Python 3.7 however, you are guaranteed that the iterator will return keys in the order they were inserted. Still this wouldn't help us here as there is no guarantee of the order in which keys are inserted into `sg_data`.

geffrak: md5 give you 20 characters... Later we concatenate two hashes to build a file name. Are you sure you want to use md5?

jfboismenu: It's all being stored in the cache folder that Toolkit uses with files named after MD5ed values, so yeah, I'm good.

jfboismenu: Famous last words. 😄

jfboismenu (approved):

PR #124, opened 2021-01-29
SG-20648 Fixes an issue with running the old style command.

pscadding:
The old style command should only be used when the app is registered with supports multiple selection. 
Before this fix, it would run the old style command if the supports_multiple_selection was present but set to false.


eshokrgozar (changes requested):
It looks like we might have two incompatible situations that we need to work through and make sure we handle both correctly. This was changed to `not None` recently to solve a problem that came up in this ticket:

Ticket: https://jira.autodesk.com/browse/SG-19840
PR: https://github.com/shotgunsoftware/tk-framework-shotgunutils/pull/122

pscadding:
After discussing this, I am closing it.

PR #128, opened 2021-07-08
SG-18781 - Set the cwd of the external_runner

geffrak:
This pull request fixes an issue where starting some DCCs from Shotgun Create fails because of the way DLLs are handled on Windows. Before this change, Shotgun Create starts external_runner.py (and then any external command) from his PWD, which is his own bin folder. The situation might end up problematic because of the way DLLs are loaded on Windows. In certain cases, the app will look at DLLs in the current directory instead of where you would expect it to load things and things end up failing in a very unexpected way.

The solution here is to always set the PWD to the home directory of the user. Why the home directory? For no reason, it's an arbitrary path that always exists and where you are unlikely to have DLLs that might break your tools. I am open to switching this to any other directory that makes sense. The other candidate in my mind is the folder that contains the argument file, but I'm afraid of what's in people's temp folders.

Some literature: https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-desktop-applications

### DIFF
--- a/python/external_config/external_command.py
+++ b/python/external_config/external_command.py
@@ -490,20 +490,23 @@ class ExternalCommand(object):
             sgtk.bootstrap.ToolkitManager.get_core_python_path(),
             args_file,
         ]
+        subprocess_cwd = sgtk.utils.sgtk.util.LocalFileStorageManager.get_global_root(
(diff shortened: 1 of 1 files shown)


jfboismenu (approved):
Looks good to me.

PR #135, opened 2021-11-18
SG-22324 [tk-framework-shotgunutils] ShotgunEntityModel intermediate item creation creates additional children, even if data handler was about to create them.

000paradox000:
This PR fixes some issues regarding ShotGrid Model when refreshing data. 

When new data arrives after a refresh due a change in the entities or filter, the children nodes were duplicated and the top items were not deleted. 

There are two main parts patched with extra validations:

- Delete item method
- Apply changes method


python/shotgun_model/shotgun_query_model.py
```diff
@@ -998,6 +1004,114 @@ def __on_sg_data_arrived(self, sg_data):
...
+                        self._log_debug("Updating model item %s" % model_item)
+                        self._update_item(model_item, data_item)
+
+            self._log_debug("...diffs applied!")
+
+        # and emit completion signal
+        self.data_refreshed.emit(len(modified_items) > 0)
+
+
+"""
```
gplsteph, line 1099: Heads up that I think you have a big text with previous code?

gplsteph (approved):
Looks good to me!

PR #143, opened 2022-12-07
Fix user settings to handle QByteArray values.

staceyoue:
* QByteArray objects are not retrieved correctly after converting to str and pickling the value
* To handle QByteArray objects, add a param to the `store` method to allow storing the raw value (without converting to str and pickling)

For context, we want to add a splitter to the Loader2 App, and save/restore the splitter state in the settings (which is a QByteArray value): https://github.com/shotgunsoftware/tk-multi-loader2/pull/102

### DIFF
--- a/python/settings/user_settings.py
+++ b/python/settings/user_settings.py
@@ -99,7 +99,7 @@ class UserSettings(object):
     ########################################################################################
     # public methods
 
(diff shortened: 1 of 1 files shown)


NorberMV (approved):
LGTM !

PR #148, opened 2023-12-27
SG-31932 Fix order when adding new items before caching

carlos-villavicencio-adsk:
When [shotgunpanel](https://github.com/shotgunsoftware/tk-multi-shotgunpanel) finds new items (notes, publishes, etc) it appends them to the end of the UI list. However, the default ordering should be the newest first. This PR attempts to fix that by specifying the index where to create new items when SG data arrives.


NorberMV (approved):
Good job, LGTM!

julien-lang (approved):
LGTM!

PR #149, opened 2024-01-24
SG-17742 Attempt to fix the crash issue with Houdini

carlos-villavicencio-adsk:
When opening the tk-multi-shotgunpanel in Houdini, the Houdini app crashes on exit. This has been [reported by the community](https://community.shotgridsoftware.com/t/houdini-sgtk-app-crash-on-close/8827).

This PR attempts to fix this by properly destroying Qt objects on close.

- Check if the current engine is Houdini so it can invoke `removeRow` instead of `takeRow` when clearing child nodes.

### DIFF
--- a/python/shotgun_model/shotgun_query_model.py
+++ b/python/shotgun_model/shotgun_query_model.py
@@ -880,11 +880,15 @@ class ShotgunQueryModel(QtGui.QStandardItemModel):
 
         # delete the child leaves
         for index in range(node.rowCount(), 0, -1):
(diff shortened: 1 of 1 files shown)


staceyoue:
@carlos-villavicencio-adsk do we know why this only happens on Houdini?

carlos-villavicencio-adsk:
@staceyoue not really, unfortunately.

And this doesn't fix 100% of the cases. But most of them in my tests.

PR #158, opened 2024-08-30
SG-36129 add handling for data processing errors

pscadding:
This was added to aid a [feature on the tk-multi-shotgunpanel ](https://github.com/shotgunsoftware/tk-multi-shotgunpanel/pull/96)where if a bad custom sort field was added it would result in a error to the console but nothing in the UI.

This change means that the `data_refresh_fail` signal will be fired if the fetched data can't be processed.

I have concerns about the impact this change may have on other tools using this framework.

### DIFF
--- a/python/shotgun_model/shotgun_model.py
+++ b/python/shotgun_model/shotgun_model.py
@@ -382,9 +382,13 @@ class ShotgunModel(ShotgunQueryModel):
 
         # construct the top level nodes
         self._log_debug("Creating model nodes for top level of data tree...")
-        nodes_generated = self._data_handler.generate_child_nodes(
-            None, root, self._create_item
-        )
+        try:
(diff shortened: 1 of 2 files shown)


carlos-villavicencio-adsk (commented):
Since we didn't have error handling on this logic before, would it make sense to make the fail silent? I mean, just log everything and let it pass? As you mentioned in your other PR, we're not sure if this can impact other dependent modules.

pscadding:
Having done more testing and discussed with @carlos-villavicencio-adsk I'm going to close this PR without merging.
We would prefer not to add risk by changing the utils here for the sake catching what will likely be a edge case scenario with the configuration of the panel app.

I still think the code change here is probably the right approach, had we chosen to go with it, so if needed in the future we could implement something similar.

PR #167, opened 2025-06-09
SG-15664 Replace `imp`

carlos-villavicencio-adsk:
This pull request refactors the `_import_py_file` function to replace the deprecated `imp` module with `importlib.util` for dynamic module loading. It also introduces a new test case to validate this functionality and adds a test fixture to support the test.

Solution based on https://github.com/shotgunsoftware/tk-photoshopcc/pull/113


python/external_config/scripts/external_runner.py
```diff
@@ -251,12 +243,13 @@ def _import_py_file(python_path, name):
     :param str name: name of py file (without extension)
     :returns: Python object
     """
-    mfile, pathname, description = imp.find_module(name, [python_path])
-    try:
-        module = imp.load_module(name, mfile, pathname, description)
-    finally:
-        if mfile:
-            mfile.close()
+    module_path = os.path.join(python_path, f"{name}.py")
```
julien-lang, line 246: CoPilot raised something here: 

image


Is this valid?

carlos-villavicencio-adsk: `os` is imported on line 11 🤷‍♂️

julien-lang (approved):

